PR #4, opened 2024-05-17
Chore/cli signatures

almutlue:
## Code signatures for cli and general code structure

This PR contains:
1. __Code signature for the cli__: 
- use, e.g. `ob --help`
- cli groups (can be renamed in `cli/main.py`
2.__repo structure__:
- very broad and simple structure suggestion for python code

__Open questions__: 
Both cli groups and the repo structure are thought as a discussion basis and suggestions, feedback is very welcome. Alternatively we could organize the cli in one file, but with this number of commands this gets long and easily messy?


DanInci (changes requested):

omni/cli/io.py
```diff
@@ -0,0 +1,124 @@
...
+):
+    """Download all or specific files for a benchmark."""
+    typer.echo(
+        f"Download {type} files for {benchmark} at stage {stage} from module {module}",
+        err=True,
+    )
+
+
+@cli.command("checksum")
+def checksum_files(
```
retogerber, line 113: Do we need a dedicated function for this? It could be done automatically when downloading.

pyproject.toml
imallona, line 1: please replace with `omni-schema = {git = "https://github.com/omnibenchmark/omni-schema.git", rev = "e3daebe9f75a6225adb674da5efaa477981baa1e"}`

imallona:
Please consider updating [the license](https://github.com/omnibenchmark/omni-py/blob/483dbfb3b860b6abced6b61d0b465768bafed917/LICENSE#L189) to something like `2024 Omnibenchmark team (our names here, including Mark)`

omni/cli/main.py
```diff
@@ -0,0 +1,82 @@
...
+            "-s",
+            help="Stage to associate the module with.",
+        ),
+    ],
+):
+    """Start a new gitlab module at a certain stage."""
+    typer.echo(
+        f"Start module {module_name} as part of {benchmark} on stage {stage}.", err=True
+    )
+    # NOTE: We probably also need a gitlab url? Not sure about module_name?
```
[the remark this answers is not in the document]
almutlue: It is about starting/setting up a new module right? So at this time point the module does not belong to a benchmark, yet and does not have an id. This function will fetch a gitlab template and set up a new project. The benchmark yaml will be used to prefill requirements like which tests to run in the cicd. The "benchmarker" will decide to include the module afterwards into the next benchmark version and then the module will get an id.

omni/cli/io.py
almutlue: @imallona do you have an opinion on this? I agree that this could be done automatically when downloading. Are there other use cases?

imallona: I agree it will be done automatically after download, but I think it would be good to have it separate/atomized and called by the download tool, when applicable. Because a checksum could be called (automated or not) to check integrity in several scenarios (after download, diff-ing, etc)

retogerber: right, I didn't think about those scenarios. Let's keep it then.

retogerber (approved):

imallona (approved):

PR #6, opened 2024-05-27
Remote Storage setup - MinIO

retogerber:
## Ticket

Prepare for IO: [list_files](https://github.com/orgs/omnibenchmark/projects/2?pane=issue&itemId=63536317), [download_files](https://github.com/orgs/omnibenchmark/projects/2?pane=issue&itemId=63536662), [checksum_files](https://github.com/orgs/omnibenchmark/projects/2?pane=issue&itemId=63536736)

## Description

Initial port from [omni_storage](https://github.com/omnibenchmark/omni_storage).

Adds classes for S3, MinIO and Swift to interact with remote storage. 
Implements: 
- [list_benchmarks](https://github.com/orgs/omnibenchmark/projects/2/views/1?pane=issue&itemId=63534208)
- [list_versions](https://github.com/orgs/omnibenchmark/projects/2/views/1?pane=issue&itemId=63534539)

**Not** yet ready to integrate, pull request was created to test remote storage with github actions secrets.

## Checklist:

## Remarks for the reviewer:


omni/io/RemoteStorage.py
```diff
@@ -0,0 +1,289 @@
...
+    Attributes:
+    - major_version (int): The major version of the storage.
+    - major_version_new (int): The new major version of the storage.
+    - minor_version (int): The minor version of the storage.
+    - minor_version_new (int): The new minor version of the storage.
+    - benchmarks (list): A list of benchmarks.
+    - versions (list): A list of versions.
+    - files (dict): A dictionary of files.
+    - benchmark (str): The current benchmark.
+    - auth_options (dict): The authentication options.
```
DanInci, line 18: I didn't quite get it from the code why the storage class has major and minor version. 
- Is this related to the underlying storage API implementation?
- Is this anyhow related to the benchmark file version?

omni/io/S3config.py
```diff
@@ -0,0 +1,101 @@
...
+                "Resource": f"arn:aws:s3:::{bucket_name}",
+            },
+            {
+                "Effect": "Allow",
+                "Principal": {"AWS": "*"},
+                "Action": "s3:GetObject",
+                "Resource": f"arn:aws:s3:::{bucket_name}/*",
+            },
+        ],
+    }
```
DanInci, line 94: Should this method that creates the readonly policy be configurable by the bucket name? This way it will be consistent with `def benchmarker_access_token_policy(benchmark)`. I think changing just the method and argument name is enough.

omni/io/S3config.py
```diff
@@ -0,0 +1,101 @@
...
+                    f"arn:aws:s3:::{benchmark}.??.?/*",
+                    f"arn:aws:s3:::{benchmark}.??.?",
+                    f"arn:aws:s3:::{benchmark}.??.??/*",
+                    f"arn:aws:s3:::{benchmark}.??.??",
+                    f"arn:aws:s3:::{benchmark}.overview/*",
+                    f"arn:aws:s3:::{benchmark}.overview",
+                    f"arn:aws:s3:::{benchmark}.test.?/*",
+                    f"arn:aws:s3:::{benchmark}.test.?",
+                    f"arn:aws:s3:::{benchmark}.test.??/*",
+                    f"arn:aws:s3:::{benchmark}.test.??",
```
DanInci, line 45: Could you write a short documentation why are there so many different bucket variations for a single benchmark? I assume it has to do with the minor and major versions, but what about `overview` and `test`?

omni/io/S3Storage.py
```diff
@@ -0,0 +1,508 @@
...
+                    "Resource": f"arn:aws:s3:::{bucket_name}",
+                },
+                {
+                    "Effect": "Allow",
+                    "Principal": {"AWS": "*"},
+                    "Action": "s3:GetObject",
+                    "Resource": f"arn:aws:s3:::{bucket_name}/*",
+                },
+            ],
+        }
```
DanInci, line 122: Can we have all policy templates separate from the code?
I think S3Config can be used to host all of them. I think this one is the same with `def bucket_readonly_policy(bucket_name):`

pyproject.toml
```diff
@@ -21,6 +21,18 @@ omni-schema = {git = "https://github.com/omnibenchmark/omni-schema.git", rev = "
...
 
+aiohttp = "^3.9.5"
+async = "^0.6.2"
+tqdm = "^4.66.4"
+requests = "^2.31.0"
+bs4 = "^0.0.2"
+lxml = "^5.2.1"
+python-dateutil = "^2.9.0.post0"
+minio = "^7.2.7"
+python-swiftclient = { version = "^4.5.0", optional = true}
```
btraven00, line 32: isn't this an extra? I think a nicer way is to declare minio or swift as 2 different extras, or complain at runtime

tests/io/test_MinIOStorage.py
```diff
@@ -0,0 +1,309 @@
+import pytest
+import re
+from omni.io.MinIOStorage import MinIOStorage
```
btraven00, line 3: pep8. also, I would do two different submodules, and leave the abc/interface at the top-level module.

tests/io/test_MinIOStorage.py
```diff
@@ -0,0 +1,309 @@
...
+else:
+    raise ValueError("No S3 credentials found")
+
+auth_options_readonly = auth_options.copy()
+del auth_options_readonly["access_key"]
+del auth_options_readonly["secret_key"]
+
+os_suffix = os.environ.get("RUNNER_OS")
+if os_suffix == "Linux" or os_suffix is None:
+    bucket_base = "test1"
```
btraven00, line 32: why the difference?

tests/io/test_MinIOStorage.py
```diff
@@ -0,0 +1,309 @@
+import pytest
+import re
+from omni.io.MinIOStorage import MinIOStorage
+import json
+import minio
+import io
+import os
+from pathlib import Path
```
btraven00, line 8: I think my high-level comment on testing here is just this:

- have you considered providing a docker-compose file for integration tests? they could be very handy for future development.

These tests look like integration tests (https://martinfowler.com/articles/practical-test-pyramid.html); I would perhaps move them to a top-level testing module; with the idea of leaving just unit-tests as module tests; perhaps with some sort of mock/doubles strategy here.

tests/io/test_files.py
```diff
@@ -0,0 +1,48 @@
...
+        auth_options = json.load(file)
+elif "S3_ENDPOINT_URL" in os.environ:
+    auth_options = {}
+    auth_options["endpoint"] = os.environ.get("S3_ENDPOINT_URL")
+    auth_options["secure"] = False
+else:
+    raise ValueError("No S3 credentials found")
+
+endpoint = (
+    f"http://{auth_options['endpoint'].replace('http://', '').replace('https://', '')}"
```
btraven00, line 19: see my comment below about forcing insecure scheme ^^

btraven00 (commented):
the feature looks complete and thoroughly tested, nice job!

a few comments I can make to improve the quality after reading the whole PR:

- Please provide a more descriptive PR title / description, and possibly a design doc (`docs/design`?).
- I would perhaps invest more time into architecting the storage providers before landing this - it will facilitate any further exploration. A simple change can be to define the methods that a storage provider MUST bring with them; and then separating different storage providers into different storage sub-modules. In this way, you can enable or disable providers according to their usefulness or maturity etc.
- Testing strategy: to improve testing during dev cycles and CI, I would perhaps move the current tests to something more reproducible than the backends you're providing here - a low hanging fruit can be to provide a different docker-compose file for testing different backend providers. then in the future you can reuse those images for integration testing in the CI...

Please take into account that all the comments I'm providing are quick suggestions, and they're probably lacking deeper context or previous discussions.  Feel free to use whatever of them is useful to you or the project! Do not hesitate in asking if something is not clear enough, or to clarify if you feel I misinterpreted the current design choices.

omni/io/README.md
```diff
@@ -0,0 +1,35 @@
...
+The to use the MinIOStorage or S3Storage class, use the following code:
+
+```
+import json
+from omni.io.MinIOStorage import MinIOStorage
+
+with open("<CONFIG>.json", "r") as file:
+    auth_options = json.load(file)
+
+ss = MinIOStorage(auth_options=auth_options, benchmark="<BENCHMARK>")
```
btraven00, line 33: I would consider renaming this variable

.github/workflows/tests.yml
```diff
@@ -61,18 +61,23 @@ jobs:
...
       # run test suite and output coverage file
       #----------------------------------------------
       - name: Run tests
         run: |
           poetry run pytest --cov=omni --cov-report=xml:./reports/coverage.xml --junitxml=./reports/junit.xml
+        env: 
+          S3_ENDPOINT_URL: ${{ secrets.S3_ENDPOINT_URL }}
+          S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
+          S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
+          RUNNER_OS: ${{ runner.os }}
```
btraven00, line 80: I do see that the runners might be configured accordingly with the system matrix that is specified in tests below, but I would try to make tests more generic, and then configure the integration tests in another layer (i.e., instantiate as many tests as platforms/runners you want from the github action, but perhaps the parallelization of the test needs to happen outside of the poetry run, as a test matrix intrinsic to the pipeline)

tests/io/test_RemoteStorage.py
```diff
@@ -0,0 +1,133 @@
+import pytest
+from omni.io.RemoteStorage import RemoteStorage
```
btraven00, line 2: relative import?

tests/io/test_RemoteStorage.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_wrong_init():
```
btraven00, line 5: for extra developer happiness, I would use some more human-friendly test names here, encoding condition and expectations (i.e., init_fails_when_foobar_is_baaz)

omni/io/S3Storage.py
```diff
@@ -0,0 +1,508 @@
...
+import logging
+from omni.io.RemoteStorage import RemoteStorage
+
+logging.basicConfig(level=logging.ERROR)
+logging.getLogger("requests").setLevel(logging.DEBUG)
+logging.getLogger("minio").setLevel(logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+
+class S3Storage(RemoteStorage):
```
btraven00, line 19: I'm confused - why is S3Storage different from MinIOStorage? Isn't MinIO a S3 implementation?

omni/io/S3Storage.py
```diff
@@ -0,0 +1,508 @@
...
+
+    def _test_connect(self):
+        try:
+            _ = list(self.client.buckets.all())
+        except Exception as e:
+            raise e
+
+    def _get_containers(self):
+        """
+        Retrieves the list of containers from the Swift storage.
```
btraven00, line 74: this is the s3 implementation isn't it?

omni/io/S3Storage.py
```diff
@@ -0,0 +1,508 @@
...
+            for version in allversions:
+                if re.match(f"(\\d+).(\\d+)", version):
+                    versions.append(version)
+                elif re.match(f"test.(\\d+)", version):
+                    other_versions.append(version)
+            self.versions = versions
+            self.other_versions = other_versions
+            return self.versions, self.other_versions
+
+            # with self.connect() as swift:
```
btraven00, line 179: delete code

omni/io/S3Storage.py
```diff
@@ -0,0 +1,508 @@
...
+            objectname (str): The name of the file.
+
+        Returns:
+            tuple: A tuple containing the file's metadata modification time and access time.
+
+        Raises:
+            requests.HTTPError: If the HTTP request to retrieve the file fails.
+        """
+        urlfile = f"{preauthurl}/{containername}/{objectname}"
+        urlfile = "http://omnibenchmark.mls.uzh.ch:9000/tb.0.1/test"
```
btraven00, line 299: why is this hardcoded here?

omni/io/files.py
```diff
@@ -1,20 +1,216 @@
...
+    if verbose:
+        print("Downloading files...")
+    tasks = [retrieve_file(url) for url in urls]
+    from tqdm.asyncio import tqdm_asyncio
+
+    headers = await tqdm_asyncio.gather(*tasks, delay=5, disable=not verbose)
+    return headers
+
+
+def get_file(url: Union[str, list[str]], verbose: bool = False):
```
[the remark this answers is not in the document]
btraven00: conceptually, the difference between `get_file` and `download_file` can be hard to tell to a module maintainer. consider renaming to be more informative.

omni/io/MinIOStorage.py
```diff
@@ -0,0 +1,457 @@
...
+        return self.benchmarks
+
+    def _create_benchmark(self, benchmark, update=True):
+        if len(self.benchmarks) == 0 or update:
+            self._get_benchmarks()
+        if benchmark in self.benchmarks:
+            raise ValueError("Benchmark already exists")
+        # create new version
+        self.client.make_bucket(bucket_name=f"{benchmark}.test.1")
+        set_bucket_public_readonly(self.client, f"{benchmark}.test.1")
```
btraven00, line 148: are test files hardcoded in this implementation?

omni/io/files.py
```diff
@@ -1,20 +1,216 @@
...
+                        break
+                    file.write(chunk)
+            return response.headers
+
+
+async def retrieve_files(urls: List[str], verbose: bool = False):
+    if verbose:
+        print("Downloading files...")
+    tasks = [retrieve_file(url) for url in urls]
+    from tqdm.asyncio import tqdm_asyncio
```
btraven00, line 107: this is cool! 
only one concern: isn't this going to mess with stdout if this is used as a library and not from the cli?

omni/cli/benchmark.py
```diff
@@ -81,6 +87,18 @@ def list_benchmarks(
 ):
     """List all available benchmarks and versions at a specific endpoint"""
     typer.echo(f"Available benchmarks at {endpoint}:")
+    benchmark_names = omni.io.files.get_benchmarks_public(endpoint)
+    benchmarks = {}
+    for benchmark in benchmark_names:
+        benchmarks[benchmark] = omni.io.files.get_benchmark_versions_public(
+            benchmark, endpoint
+        )
+    for key, value in sorted(benchmarks.items()):
```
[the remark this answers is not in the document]
retogerber: Agreed, this can be simplified since 'value' is expected to be a (potentially empty) list.

omni/io/RemoteStorage.py
retogerber: I added some description of what the versioning is for here: [omni/io/README.md](https://github.com/omnibenchmark/omni-py/blob/670de168c986368ac5aebe17b591920408220029/omni/io/README.md). 
In short, the versioning is purely for the benchmarking to be able to have different versions of a benchmark. The versions are specified in the benchmark config file.

omni/io/files.py
```diff
@@ -1,20 +1,216 @@
-"""Functions to manage file handling"""
+"""Functions to manage files"""
 
+import asyncio
+import aiohttp
+from urllib.parse import urlparse
+import re
+from pathlib import Path
+from omni.io.utils import get_storage, md5
+from omni.sync import get_bench_definition
```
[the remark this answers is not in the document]
retogerber: Thanks, yes good point.

omni/io/RemoteStorage.py
```diff
@@ -0,0 +1,289 @@
...
+
+        Args:
+            benchmark: The benchmark object to be created.
+
+        Returns:
+            None
+        """
+        NotImplementedError
+
+    def _get_versions(self, update: bool = True, readonly: bool = False):
```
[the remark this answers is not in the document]
retogerber: Thanks, wasn't aware of this. Will refactor accordingly.

tests/io/test_RemoteStorage.py
retogerber: good point, will change that.

omni/io/S3Storage.py
retogerber: This is indeed unnecessary, I tried to get the same outcome with using boto3 than with minio and kept the result. They don't always are exactly the same, for minio I encountered one situation where additional access policy needed to be set to get the same result. Anyway, I removed the S3Storage class with [423d512167df5233f283b8ec4673c3ac0b9a76a4](https://github.com/omnibenchmark/omni-py/commit/423d512167df5233f283b8ec4673c3ac0b9a76a4)

omni/io/MinIOStorage.py
retogerber: yes and no, see https://github.com/omnibenchmark/omni-py/blob/io/omni/io/README.md . In short this is a bucket intended to contain a test benchmark for testing modules.

omni/io/RemoteStorage.py
```diff
@@ -0,0 +1,289 @@
+"""Base class for remote storage."""
+from typing import Union, List, Dict
+
+
+class RemoteStorage:
```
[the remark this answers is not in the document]
retogerber: Thanks for the suggestion, that was pretty much as I envisioned it. Changed the ReomteStorage class to be an ABC.

tests/io/test_MinIOStorage.py
retogerber: that was to fix failures in testing with github actions which run concurrently. If the same bucket names where used it led to mix ups. The testing changed to using testcontainers which should isolate the tests more.

tests/io/test_MinIOStorage.py
retogerber: I switched now to using testcontainers (see above) which should help for future development. Not sure how to globaly structure tests in the future, but for now I think it might make sense to leave them as they are now.

omni/io/S3config.py
retogerber: I prepared an overview here: https://github.com/omnibenchmark/omni-py/blob/io/omni/io/README.md
in short the `overview` is the bucket to retrieve public info about versions and `test` is/are buckets intended for unit tests of benchmarks.

omni/io/S3Storage.py
retogerber: Agreed, moved all to the file S3config.py

omni/io/S3config.py
retogerber: I see what you mean, but here I want to set the policy for a single version of a benchmark (i.e. `benchmark.0.1`) and not for all possible benchmark versions/buckets. in `benchmarker_access_token_policy` the access policy is intended to give access to all possible bucket names for a benchmark.

omni/io/S3config.py
```diff
@@ -0,0 +1,101 @@
+def tests_bucket_policy():
```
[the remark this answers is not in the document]
retogerber: Good point. With updated testing, this is not needed anymore.

.github/workflows/tests.yml
retogerber: Switched to `testcontainers` which makes this obsolete, still maybe some changes will be necessary in the future.

omni/io/S3config.py
DanInci: This is much more clear now. And the visuals really helps, thanks and good job documenting it !

omni/io/files.py
```diff
@@ -1,20 +1,205 @@
...
+    benchmark: str,
+    type: str,
+    stage: str,
+    module: str,
+    file_id: str,
+    version: str,
+    verbose: bool = False,
+):
+    """Compare md5 checksums of available files for a certain benchmark, version and stage with local versions"""
+
```
DanInci, line 149: Its cool to have this as a separate feature / cli command as well.
I was wondering if it would be useful to also checksum implicitly when downloading files?

retogerber: This is already implemented, see: https://github.com/omnibenchmark/omni-py/blob/b72669a9f0be97686cfb6a4341a9b4b6611f1bba/omni/io/files.py#L113

DanInci (approved):

PR #8, opened 2024-07-29
One bucket one benchmark storage

retogerber:
## Description

Change remote storage backend from "one-version-one-bucket" to "one-benchmark-one-bucket". Instead of creating a new bucket for each new benchmark version a single version-controlled bucket is used which should solve some issues (storage overhead, potential existing buckets) at the cost of some complexity. 

## Checklist:

## Remarks for the reviewer:

New design description [here](https://github.com/omnibenchmark/omni-py/blob/one_bucket_one_benchmark_storage/omni/io/README.md) and example [here](https://github.com/omnibenchmark/omni-py/blob/c3278bb91d9027286a7dae8f9dd010df76da8d49/examples/storage_usage.py).


omni/io/files.py
```diff
@@ -131,8 +60,42 @@ def download_files(
...
+        minio_auth_options_public = {
+            "endpoint": "https://omnibenchmark.mls.uzh.ch:9000",
+            "secure": False,
+        }
+        bench_yaml = {
+            "auth_options": minio_auth_options_public,
+            "storage_type": "minio",
+        }
+        benchmark = "testversioning"
+        version = "0.1"
```
DanInci, line 81: These are the current specs of the benchmark model. Do we miss something? What are the auth options in this context? 
```
id: Benchmark_001
description: simple benchmark, somewhat explicit, simple params
version: 1.0
benchmarker: "John Doe at Robinsons lab, john.doe@uzh.ch"
storage: https://storage.github.com/
storage_api: S3
benchmark_yaml_spec: 0.01
```

tests/io/test_list_versions.py
```diff
@@ -1,27 +0,0 @@
...
-    with TmpMinIOStorage(minio_testcontainer) as tmp:
-        _ = MinIOStorage(auth_options=tmp.auth_options, benchmark="bm")
-        ocb.list_versions("bm", tmp.auth_options_readonly["endpoint"])
-        out, err = capfd.readouterr()
-        assert len(out) > 0
-
-        with pytest.raises(requests.exceptions.HTTPError):
-            ocb.list_versions(
-                "not_existing_benchmark", tmp.auth_options_readonly["endpoint"]
-            )
```
DanInci, line 27: Shouldn't this CLI test still exist?

omni/io/MinIOStorage.py
```diff
@@ -4,452 +4,378 @@
...
+                and "access_key" in self.auth_options.keys()
+                and "secret_key" in self.auth_options.keys()
+            )
+            tmp_auth_options = self.auth_options.copy()
+        try:
+            return minio.Minio(**tmp_auth_options)
+        except Exception as e:
+            url = urlparse(tmp_auth_options["endpoint"])
+            tmp_auth_options["endpoint"] = url.netloc
+            return minio.Minio(**tmp_auth_options)
```
DanInci, line 221: Instead of using assertion, I would introduce some custom errors & error handling for storage

omni/io/MinIOStorage.py
```diff
@@ -4,452 +4,378 @@
...
+                            if is_valid_version(k):
+                                tags_filt[k] = w
+            di[uq][v] = {}
+            di[uq][v]["tags"] = tags_filt
+            tmpinds2 = [i for i, tvers in enumerate(tv) if tvers == v]
+            di[uq][v]["size"] = ts[tmpinds2[0]]
+            di[uq][v]["last_modified"] = tlm[tmpinds2[0]]
+            di[uq][v]["is_delete_marker"] = tdl[tmpinds2[0]]
+            di[uq][v]["etag"] = etl[tmpinds2[0]]
+    return di
```
DanInci, line 68: This logic is hard to read through, but if I understand it correctly it aggregates unique objects (objects_ls) by their name and creates a dictionary from `name` -> `object` containing metadata of the aggregated objects.

One suggestion, wouldn't it be more straightforwards to group by followed by a map?
Something like:
```python

# Group listed objects by their name
grouped_objects = groupby(object_ls, key=lambda o: o.object_name)

di = {}

# Iterate over object groups
for object_name, objects in grouped_objects:
    objects = list(objects)
    di[object_name] = {}
    
    # Extract attributes for the objects group
    version_ids = [o.version_id for o in objects]
    sizes = [o.size for o in objects]
    last_modified = [o.last_modified for o in objects]
    is_delete_markers = [o.is_delete_marker for o in objects]
    etags = [o.etag for o in objects]
    
    # Iterate to all versions and create a dictionary entry
    for i, v in enumerate(version_ids):
        tags_filt = {}
        if not readonly and not is_delete_markers[i]:
            tags = client.get_object_tags(benchmark, object_name, version_id=v)
            if tags is not None:
                tags_filt = {k: w for k, w in tags.items() if is_valid_version(k)}
        
        di[object_name][v] = {
            "tags": tags_filt,
            "size": sizes[i],
            "last_modified": last_modified[i],
            "is_delete_marker": is_delete_markers[i],
            "etag": etags[i]
        }
```

DanInci (commented):
Overall the logic looks good to me.
Here are some thoughts / suggestions, with different levels of priority (ordered them highest to lowest):
- would be really useful to add type hints in MinIOStorage, I spent too much time trying to understand what the arguments are.
- with these changes there's a lot more logic to efficiently manage the storage.  I think it should be properly documented, maybe even split in more compact helper functions, so that we don't run the risk of forgetting what happens.
- we should think of a unifying mechanism for error handling throught the project, also for logging, because assertion won't work, and errors need to be propagated and handled by the CLI layer, so they are intuitive to the user.

omni/io/MinIOStorage.py
retogerber: Thanks, agree, wasn't very clear and no type hints. Refactored this.

omni/io/RemoteStorage.py
```diff
@@ -136,14 +126,23 @@ def _parse_current_version(self, version: Union[None, str] = None) -> Version:
...
         if version is None:
-            version_out = max(self.versions)
+            if self.versions == []:
+                version_out = Version("0.1")
+            else:
+                version_out = max(self.versions)
+                if version_out.minor > 98:
+                    raise ValueError(
+                        f"Minor Version {version_out.minor} cannot be incremented. Please increment the major version."
+                    )
```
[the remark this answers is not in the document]
retogerber: There was one with the previous storage implementation, now it is not needed anymore. Removed this.

retogerber:
> Overall the logic looks good to me. Here are some thoughts / suggestions, with different levels of priority (ordered them highest to lowest):
> 
> * would be really useful to add type hints in MinIOStorage, I spent too much time trying to understand what the arguments are.
> * with these changes there's a lot more logic to efficiently manage the storage.  I think it should be properly documented, maybe even split in more compact helper functions, so that we don't run the risk of forgetting what happens.
> * we should think of a unifying mechanism for error handling throught the project, also for logging, because assertion won't work, and errors need to be propagated and handled by the CLI layer, so they are intuitive to the user.

Thanks for the feedback. 
I added some more type hints and documentation which hopefully helps in understanding. 
I agree that we should have some global consistent way of handling errors. But I think this might be something to implement once the integration of the different parts is complete.

DanInci (approved):

PR #12, opened 2024-08-28
streamline poetry/pip installs and testing (e.g. boto3 problems)

imallona:
## Ticket

No ticket, boto3 installation problems. Attempting to get it to work with both poetry and pip, without reinstalling (forcefully) with pip.

## Description

Update testing structure, so it's on pull request and environments are hackily tuned.

## Checklist:

## Remarks for the reviewer:

Notice now all checks are run on pull, regardless of which paths were modified.

Also, mac and linux software tests run jointly. Using `tests-environment.yml` or `mac-tests-environment.yml` and micromamba. No poetry.

Also, other tests run jointly (on a different testing environment, using poetry).


.github/workflows/tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]
```
DanInci, line 15: Hmm, do we want to remove all windows tests?
How about the idea we discussed yesterday where specific tests are excluded to be run from windows?

imallona: oh, this was to avoid https://github.com/omnibenchmark/omnibenchmark/actions/runs/10596355333

imallona: I've indeed started to separate software-related from other tests by using https://github.com/omnibenchmark/omnibenchmark/blob/5b7701dbece2750dd46da3656a8b84a54f4345a4/.github/workflows/tests.yml#L86 . Because we indeed had some skipped tests and duplications. This I did by adding the second line in this snippet:

```
          poetry run pytest --cov=omni --cov-report=xml:./reports/coverage.xml --junitxml=./reports/junit.xml \
              tests/benchmark tests/cli tests/workflow tests/io  ## <<---
```

which breaks on windows

imallona: not due to this https://github.com/omnibenchmark/omnibenchmark/commit/c2754928c544f38d37e1933ca208889df76bd164 , it breaks with a normal `poetry run pytest too`. I'd skip the windows tests for now, we have more urgent matters

imallona: this is the last successful run https://github.com/omnibenchmark/omnibenchmark/actions/runs/10578177745/workflow

DanInci: Alright sounds good to me

DanInci (approved):

PR #17, opened 2024-09-13
Switch from typer to click to speed up

imallona:
## Checklist:

## Remarks for the reviewer:

@retogerber , what do you think, time-wise? worth pursuing? I've only added 'software' signatures

Major important stuff:
- @DanInci : mind I've butchered your run benchmark and run method, note I'm not adding the 'example' 'remote' 'all' unimplemented capabilities, and that I also modified the sys.exits on error
- All: I'm adding the module imports within commands, so the CLI is faster when just exploring its commands/arguments. This is a two sided sword...
- The single-file-click-cli is rather dirty
- Only `software` and `run` commands are included, Reto will add his


pyproject.toml
```diff
@@ -20,7 +20,9 @@ classifiers = [
 python = "^3.12"
 #setuptools="^72.1.0"
 
-typer = {extras = ["all"], version = "^0.12.3"}
+# typer for testing only
+typer = {extras = ["all"], version = "^0.12.3"} 
+click="^8.1.7"
 typing-extensions = "^4.11.0"
```
DanInci, line 26: Typer should be added as a test dependency then.

DanInci (commented):
Migration to click looks good, although why not have separete files for the CLI interface, like before, with run, software and storage separated?

retogerber:
> Migration to click looks good, although why not have separete files for the CLI interface, like before, with run, software and storage separated?

I'll give it a try migrating to multiple files.

retogerber (approved):

imallona:
Thanks @retogerber, I think this is ready to merge :muscle:

PR #19, opened 2024-09-19
Software backend updates to align to Daniel's

imallona:
Food for thought re: the software backends.

Mainly, the issue is that module names are not files but just words (e.g. R/latest) checked against an index typically stored as a set of lua module files inside the $MODULEPATH dir. So some validations do not work. 

We could address this in several ways:
- Check whether the module file (lua, inside the $MODULEPATH) exists, which only will happen if the module is available
- Check whether the module is available/loadable instead (e.g. using `module` commands, for instance `module avail`)
- Check whether the easyconfig is available instead (this proposal)

As for the easyconfig, they're normally not full paths neither, but rather files within the [robot search path](https://docs.easybuild.io/using-easybuild/#robot_search_path). I understand the advantages of having them in `env` folders next to the benchmarking YAML. So I've tried to circunvent this updating the searchpath / robotspath to both the "default" and the `os.path.getcwd()` but this is not ideal.

## Checklist:

### DIFF
--- a/.github/workflows/software_tests.yml
+++ b/.github/workflows/software_tests.yml
@@ -178,7 +178,9 @@ jobs:
 
           mkdir -p "$HOME"/.local/easybuild/modules/all
           module use "$HOME"/.local/easybuild/modules/all
+          #module use "$HOME"/micromamba/envs/omnibenchmark/modulefiles/all
 
+          export MODULEPATH="$HOME"/micromamba/envs/omnibenchmark/modulefiles/all:$MODULEPATH
           export MODULEPATH="$HOME"/.local/easybuild/modules/all:$MODULEPATH
           
           export PYTHONPATH=${PYTHONPATH}:$LMOD_DIR/../init
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04] # [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest] # [ubuntu-22.04, macos-latest]
         python-version: ["3.12"]
         poetry-version: ["1.8.0"]
         r-version: ["4.3.3"]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest] # [ubuntu-latest, macos-latest]
         python-version: ["3.12"]
         poetry-version: ["1.8.0"]
         r-version: ["4.3.3"]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest] # [ubuntu-latest, macos-latest]
+        os: [ubuntu-22.04] # [ubuntu-latest, macos-latest]
         python-version: ["3.12"]
         poetry-version: ["1.8.0"]
         r-version: ["4.3.3"]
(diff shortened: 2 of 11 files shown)


imallona:
@retogerber ready for review thanks. (Assuming tests work). If tests are green, could you please also check that both the envmodules prep/easyconfig building and the envmodules-based benchmark run works? Thanks!

```bash
echo $MODULEPATH ## out of curiosity, should be something like the ones below

# just in case we export it
export MODULEPATH=$HOME/micromamba/envs/omnibenchmark/modulefiles/Core:$HOME/micromamba/envs/omnibenchmark/modulefiles/Core/all

ob software envmodules prepare -b tests/data/Benchmark_001.yaml

# should list the two zlib versions
module avail

# should run reporting it's using the envmodules (the two zlib ones)
ob run benchmark -b tests/data/Benchmark_001.yaml --local
```

retogerber:
> @retogerber ready for review thanks. (Assuming tests work). If tests are green, could you please also check that both the envmodules prep/easyconfig building and the envmodules-based benchmark run works? Thanks!
> 
> ```shell
> echo $MODULEPATH ## out of curiosity, should be something like the ones below
> 
> # just in case we export it
> export MODULEPATH=$HOME/micromamba/envs/omnibenchmark/modulefiles/Core:$HOME/micromamba/envs/omnibenchmark/modulefiles/Core/all
> 
> ob software envmodules prepare -b tests/data/Benchmark_001.yaml
> 
> # should list the two zlib versions
> module avail
> 
> # should run reporting it's using the envmodules (the two zlib ones)
> ob run benchmark -b tests/data/Benchmark_001.yaml --local
> ```
Unfortunately does not work for me
```
$ micromamba activate 
$ micromamba create -n omnibenchmark
Found conda-prefix at '/home/retger/micromamba/envs/omnibenchmark'. Overwrite?: [y/N] y
Empty environment created at prefix: /home/retger/micromamba/envs/omnibenchmark
$ micromamba activate omnibenchmark
$ micromamba install -f test-environment.yml
...
$ echo $MODULEPATH ## out of curiosity, should be something like the ones below
/home/retger/micromamba/envs/omnibenchmark/modulefiles/Linux:/home/retger/micromamba/envs/omnibenchmark/modulefiles/Core:/home/retger/micromamba/envs/omnibenchmark/lmod/lmod/modulefiles/Core
$ export MODULEPATH=$HOME/micromamba/envs/omnibenchmark/modulefiles/Core:$HOME/micromamba/envs/omnibenchmark/modulefiles/Core/all 
$ ob software envmodules prepare -b tests/data/Benchmark_001.yaml
...
$ module avail
Lmod has detected the following error:  module avail is not possible. MODULEPATH is not set or not set with valid paths.

```

imallona:
yeah, thanks, it's going to be hard to harmonize, easybuild gets its own (~/.local) defaults even if installed with micromamba, micromamba has priority for robotspath (?), and then lmod does something else. In fact by default the $MODULEPATH is a blob of very different things in most systems I've tried...

I've asked Mark and for him is even funnier on a Mac. I'll document all this and perhaps provide more `check` features to make sure the modulepath exists, is clear, read by lmod etc

```
(omnibenchmark) mark@MLS-M-MAROBI omnibenchmark % eb --show-config
#
# Current EasyBuild configuration
# (C: command line argument, D: default value, E: environment variable, F: configuration file)
#
buildpath      (D) = /Users/mark/.local/easybuild/build
containerpath  (D) = /Users/mark/.local/easybuild/containers
installpath    (D) = /Users/mark/.local/easybuild
repositorypath (D) = /Users/mark/.local/easybuild/ebfiles_repo
robot-paths    (D) = /Users/mark/micromamba/envs/omnibenchmark/easybuild/easyconfigs
sourcepath     (D) = /Users/mark/.local/easybuild/sources
(omnibenchmark) mark@MLS-M-MAROBI omnibenchmark % echo $MODULEPATH
/opt/homebrew/Cellar/lmod/8.7.49/modulefiles/Darwin:/opt/homebrew/Cellar/lmod/8.7.49/modulefiles/Core

```

DanInci (approved):

PR #21, opened 2024-09-24
Plot computational graph & export benchmark topology using Mermaid format

DanInci:
## Ticket
- No referenced ticket.

## Description

I've added 2 new CLI commands which can be used to either plot the computational graph using GraphViz or to export the benchmark topology using Mermaid format.

**1. Plot the computational graph**
```
ob info computational -b /tests/data/Benchmark_001.yaml
```

Output `{benchmark_name}.png` that looks like:
Benchmark_001

**2. Export benchmark topology**
```
ob info topology -b /tests/data/Benchmark_001.yaml
```
The following stdout:
```mermaid
---
title: Benchmark_001
---
flowchart LR
	classDef param fill:#f96
	subgraph data
		D1
		D2
	end
	subgraph process
		P1
		D1 --> P1
		D2 --> P1
		P2
		D1 --> P2
		D2 --> P2
	end
	subgraph methods
		M1
		D1 --> M1
		D2 --> M1
		P1 --> M1
		P2 --> M1
		M2
		D1 --> M2
		D2 --> M2
		P1 --> M2
		P2 --> M2
	end
	subgraph metrics
		m1
		M1 --> m1
		M2 --> m1
		m2
		M1 --> m2
		M2 --> m2
		m3
		M1 --> m3
		M2 --> m3
	end
	subgraph params_P1
		215db671-7dd2-481c-9de7-01b8a54eb4ac['-a 0', '-b 0.1']
		17ece903-0b28-415c-8960-8426bcf2f3a6['-a 1', '-b 0.1']
	end
	params_P1:::param --o P1
	subgraph params_P2
		725d641d-5405-481a-89a5-4d3018e88807['-a 0', '-b 0']
		63330392-0315-4bc7-9ca5-e849ebe5e4a5['-a 1', '-b 0.1']
	end
	params_P2:::param --o P2
	subgraph params_M2
		73d26c2a-47d2-462d-b052-3fc39997610e['-d1', '-e 1']
		48136303-2e37-4b2a-b403-2ed30ca57d79['-d1', '-e 2']
	end
	params_M2:::param --o M2
```

## Checklist:


omni/benchmark/dag.py
```diff
@@ -150,22 +150,30 @@ def compute_stage_order(stage_dag: nx.DiGraph) -> List:
     return topological_order
 
 
-def plot_graph(
-    g, output_file, scale_factor=1.0, node_spacing=0.1, figure_size=(12, 12)
+def export_to_figure(
+    g, scale_factor=1.0, node_spacing=0.1, figure_size=(12, 12), title=None
```
imallona, line 154: curiosity: is the node/edge size "reactive" to adapt to the `figure_size` so complex layouts are not crowded?

imallona (dismissed):
Looks good, thanks! I'm aware the macos tests are due to the environment being broken and not the tests themselves

imallona:
Thanks for fixing the tests. I've tested it again. Looks good to me, I'd merge even without @retogerber 's feedback.

omni/benchmark/dag.py
DanInci: In this outdated format, it was not. Basically there was very few logic involved in plotting the computational graph.
I'll post below some details about the new plotting design and my experimentation journey.

DanInci:
Experimented a bit with graph layouting and came up with a new, somewhat scalable design for plotting the computational graph of the benchmarking system.

I've tried different layout algorithms, from circular, spring models, force directed placement, but the only one that makes sense, given that we are dealing with computational pipelines, is the hierarchical layout algorithm.

Problem is that I could only find that algorithm provided in the `graphviz` software, and use it with a python wrapper `pygraphviz`. This means we'll have to add graphviz as a requirement for the package, or alternatively, I added it as an extra requirement just for the plotting command / functionality. I'm not sure how willing are we to add this?

Here's how it looks:
- layout is top to bottom
- nodes are coloured by their corresponding computational stage
- node sizes are reactive to the number of nodes
- figure size is reactive to the number of nodes
- labels are positioned below nodes

Let me know what you think / if you have suggestions.

peiying_with_easyconfigs
Benchmark_001

And I've also experimented with scaling the plot using a simulated benchmark between 10 and 100 nodes.
small_graph
medium_graph
large_graph

DanInci:
One alternative that I can see is to have plotting with different layouts available, some of them requiring graphviz, while other work out of the box using `networkx`

For a list of available layout algorithms check this:
https://networkx.org/documentation/stable/reference/drawing.html

retogerber:
Using a multipartite layout could maybe be an option: https://networkx.org/documentation/stable/auto_examples/drawing/plot_multipartite_graph.html#multipartite-layout

For example:
```
# from omni.benchmark import Benchmark
import networkx as nx
import matplotlib.pyplot as plt


# bm = Benchmark("Benchmark_001.yaml")
# G= bm.G
def draw_dag(G):
    nodes = G.nodes()
    tt = list(nodes.keys())[0]

    stage_order = ["data", "process", "methods", "metrics"]
    node_relabel_dict = {}
    for tt in nodes:
        node_relabel_dict[tt] = tt.get_id()
        G.add_node(tt, stage=f"{[i for i,st in enumerate(stage_order) if st in tt.stage.id][0]}-{tt.stage.id}")
    G = nx.relabel_nodes(G, node_relabel_dict)

    pos = nx.multipartite_layout(G, subset_key="stage", align="horizontal")
    subset_color={"0-data":"blue", "1-process":"green", "2-methods":"red", "3-metrics":"yellow"}
    color = [subset_color[data["stage"]] for v, data in G.nodes(data=True)]
    fig, _ = plt.subplots(1)
    nx.draw(G, pos, node_color=color, with_labels=True)
    return fig
```

image

Would need some more refinement of course and not sure about scalability.

DanInci:
Haven't tried multipartite before, but looks really promising !! Definitely needs some tweaks, but I could experiment more with it next week.

bio-traven:
the networkx layouts are know for being a bit poor to the eye.

just a wild idea, but for graph or a certain complexity, perhaps an additional option might be using something on the browser (like d3). the idea being that user can adjust different layouts interactively (more like the gephi experience). this also would allow to collapse nodes in boxes or clusters (and expand them interactively).

one "quick" way of trying this out plotly would be https://plotly.com/python/network-graphs/

omni/benchmark/benchmark.py
```diff
@@ -114,10 +115,61 @@ def get_available_parameter(self, module_id: str):
...
+        fig = dag.export_to_figure(
+            self.G,
+            layout_design="hierarchical",
+            title=self.get_benchmark_name(),
         )
+        fig.savefig(output_file)
+
+    def export_to_mermaid(self, show_params: bool = True) -> str:
+        # Initialize the mermaid diagram syntax
+        mermaid_diagram = f"---\n"
```
bio-traven, line 128: not sure if this will need to evolve in the future or not, but if complexity increases I would consider jinja or similar templating libs.

omni/benchmark/dag.py
```diff
@@ -150,22 +153,87 @@ def compute_stage_order(stage_dag: nx.DiGraph) -> List:
     return topological_order
 
 
-def plot_graph(
-    g, output_file, scale_factor=1.0, node_spacing=0.1, figure_size=(12, 12)
-):
-    layout = nx.circular_layout(g, scale=scale_factor)
+def export_to_figure(G, layout_design="hierarchical", title=None):
```
bio-traven, line 156: perhaps use a constant for "hierarchical"

omni/benchmark/dag.py
```diff
@@ -150,22 +153,87 @@ def compute_stage_order(stage_dag: nx.DiGraph) -> List:
...
-    plt.figure(figsize=figure_size)
+        layout = pygraphviz_layout(G, prog="neato")
+    else:
+        raise ValueError(
+            "Graph can only be exported using the `spring` or `hierarchical` layouts."
+        )
+
+    # Dynamically scale the figure size based on node count
+    nodes_count = len(G.nodes)
+    figure_size = (max(15, nodes_count // 4), max(15, nodes_count // 4))
```
bio-traven, line 170: constants will make this easier to modify

bio-traven (commented):
nice feature! the optional graph is a good compromise in my view. the layouts can be beautified later on, but this seems a good start.

imallona:
Can you update the PR to attempt to merge to `dev` and not to `main`? So we group multiple features (PRs) into releases/version bumps?

DanInci:
@imallona just did that. Let me first wrap up the [other PR](https://github.com/omnibenchmark/omnibenchmark/pull/29) that removes the dependency on graphviz, and merge it into this branch.

imallona (approved):

retogerber (approved):

PR #22, opened 2024-09-25
Do not require input_dir when the node is an entrypoint module

DanInci:
## Description

Instead of having to always add the `--input_dir` flag when running an individual module, skip it when we are referring to entrypoint modules. The default input directory is assumed to be the current work directory.

The following command change:
```
ob run module --benchmark tests/data/Benchmark_001.yaml --module D1 --input_dir .
```
to
```
ob run module --benchmark tests/data/Benchmark_001.yaml --module D1
```

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/.github/workflows/software_tests.yml
+++ b/.github/workflows/software_tests.yml
@@ -101,7 +101,7 @@ jobs:
           pytest -v -x --show-capture=stderr \
               --splits 2 --group ${{ matrix.test_group }} --splitting-algorithm=least_duration \
               cli_conda_tests.py cli_singularity_tests.py cli_envmodules_tests.py \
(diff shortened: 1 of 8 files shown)


imallona:
I wonder whether the failing tests might have to do with the runner caching...

imallona:
nope clearing the cache is not helping, can you @DanInci check the failing tests?

imallona:
Works perfectly, thanks! Can you:
1. update the click echo `Running module on a dataset provided in a custom directory` in this case, since it's not provided in a custom directory
2. update the tutorial so the input directory is not provided anymore?

imallona (approved):

PR #38, opened 2024-11-14
Fix envmodules-aware software testing (CI/CD)

imallona:
Fixing envmodules test-related software tests


omni/workflow/snakemake/snakemake.py
```diff
@@ -31,6 +31,7 @@ def run_workflow(
         dryrun: bool = False,
         work_dir: Path = Path(os.getcwd()),
         backend: SoftwareBackendEnum = SoftwareBackendEnum.host,
+        modulepath=os.environ["MODULEPATH"],
         **snakemake_kwargs,
     ) -> bool:
```
DanInci, line 36: This default value is going to fail during runtime if the `MODULEPATH` does not exist.
Also the overriding option should be added to `def run_module_workflow` also, and also in the workflow.py interface and docstrings.

omni/workflow/snakemake/snakemake.py
```diff
@@ -57,6 +58,8 @@ def run_workflow(
             snakefile, cores, update, dryrun, backend, work_dir, **snakemake_kwargs
         )
 
+        os.environ["MODULEPATH"] = modulepath
+
```
DanInci, line 62:
```
        if modulepath:
            os.environ["MODULEPATH"] = modulepath
```

omni/workflow/snakemake/snakemake.py
DanInci: modulepath=os.environ.get("MODULEPATH", None)

DanInci (changes requested):

PR #44, opened 2024-12-17
Add poetry-dynamic-versioning & logging

DanInci:
## Ticket

No reference ticket.

## Description

Added the following features:
- [x] Dynamic versioning managed by [poetry-dynamic-versioning ](https://pypi.org/project/poetry-dynamic-versioning/), using a semvar style.
- [x] `ob --version` which outputs the current version of the cli
- [x] Uniform logging + configuration and management across the whole codebase, no longer using click.echo, rather logger.log
- [x] Black pre-commit hook 
- [x] Redirect subprocess stdout and stderr to separate files.

## Checklist:
- [x] My code follows the code style of this project.
- [x] My CLI method respects the signature defined in the task.
- [x] I have documented the CLI method accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a CHANGELOG entry.

PR #45, opened 2025-01-09
Add `benchmark` directives and aggregate `benchmark` (metrics) results

imallona:
## Ticket

No ticket

## Description

Updated the workflow engine to make use of `benchmark` (rule) directives. Wrote a metric gatherer for those that gathers, in python/pandas, performance metric (`benchmark` directive outputs) files and aggregates them into a single tsv.

## Checklist:

## Remarks for the reviewer:

- This is not meant to be merged as-is, but, instead, is food for thought on how to aggregate metrics `on success`.
- Gatherers/collectors are relevant for other metrics too, as in https://github.com/omnibenchmark/omni-schema/pull/11
- Past commits also had `log` directives added. I've removed them because these should be aligned to our custom logger (#44).


omni/workflow/snakemake/scripts/parse_performance.py
```diff
@@ -0,0 +1,74 @@
...
+            with open(param_file_path) as fh:
+                reader = csv.reader(fh, delimiter="\t")
+                for row in reader:
+                    res = "%s %s %s %s %s;" % (
+                        res,
+                        triple[0],
+                        triple[1],
+                        triple[2],
+                        row[0].strip(),
+                    )
```
DanInci, line 67: I suggest an improvement for readability here.
Introduce a temporary class that wraps the tokenized file path, which contains: `stage`, `module`, `param`.
Working with triples and accessing accessing them by index in a vector can become unreadable to someone that has no context.

imallona:
Closing and opening a slim version - without collection. So the future metrics collector will handle the aggregation #54

Issue #46, opened 2025-01-13
Failure to run benchmark with `snakemake>8.24`


imallona:
As reported in https://github.com/omnibenchmark/clustering_example/pull/2 `ob run benchmark -b Clustering.yml --local` runs with dependency `snakemake==8.24.1` but fails with `snakemake==8.25.5` and `snakemake==8.27.1`.

Trace `snakemake==8.25.5`:

<details>

```
(omnibenchmark) [imallona@MLS-L-IZMA clustering_example]$ ob run benchmark  -b Clustering.yaml --local
Benchmark YAML file integrity check passed.
Running benchmark...
Assuming unrestricted shared filesystem usage.
host: MLS-L-IZMA
Building DAG of jobs...
Your conda installation is not configured to use strict channel priorities. This is however important for having robust and correct environments (for details, see https://conda-forge.org/docs/user/tipsandtricks.html). Please consider to configure strict priorities by executing 'conda config --set channel_priority strict'.
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job                        count
-----------------------  -------
all                            1
data_clustbench_param_0        1
data_clustbench_param_1        1
data_clustbench_param_2        1
total                          4

Select jobs to execute...
Execute 1 jobs...

[Mon Dec 16 11:14:55 2024]
localrule data_clustbench_param_2:
    output: out/data/clustbench/param_2/clustbench.data.gz, out/data/clustbench/param_2/clustbench.labels0.gz
    jobid: 3
    reason: Missing output files: out/data/clustbench/param_2/clustbench.labels0.gz, out/data/clustbench/param_2/clustbench.data.gz
    wildcards: stage=data, module=clustbench, params=param_2, dataset=clustbench
    resources: tmpdir=/home/imallona/tmp

Activating conda environment: .snakemake/conda/19d5081c77a4cb3f39a74cad33393bb9_
Activating conda environment: .snakemake/conda/19d5081c77a4cb3f39a74cad33393bb9_
Traceback (most recent call last):

  File "/home/imallona/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/snakemake/cli.py", line 2156, in args_to_api
    dag_api.execute_workflow(

  File "/home/imallona/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/snakemake/api.py", line 595, in execute_workflow
    workflow.execute(

[...]

PR #48, opened 2025-01-14
Fix failure to run benchmark with snakemake > 8.24

DanInci:
## Ticket

Fixes the following [Issue](https://github.com/omnibenchmark/omnibenchmark/issues/46).

## Checklist:

### DIFF
--- a/omni/workflow/snakemake/rules/rule_node.smk
+++ b/omni/workflow/snakemake/rules/rule_node.smk
@@ -31,12 +31,16 @@ def _create_initial_node(benchmark, node):
             dataset=module_id
         output:
             formatter.format_output_templates_to_be_expanded(node)
+
+        # Snakemake 8.25.2 introduced changes that no longer allow None for environment values
+        # Hence we provide alternatives for `conda`, `envmodules`, `container` which do not exist, although it will not affect the normal flow
+        # See https://github.com/snakemake/snakemake/releases/tag/v8.25.2
         conda:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.conda)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.conda) or "conda_not_provided.yml"
         envmodules:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.envmodules)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.envmodules) or "module/not_provided/0.0.0"
         container:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.apptainer)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.apptainer) or "container_not_provided.sif"
         params:
             repository_url = repository_url,
             commit_hash = commit_hash,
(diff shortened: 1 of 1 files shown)


omni/workflow/snakemake/rules/rule_node.smk
```diff
@@ -32,11 +32,11 @@ def _create_initial_node(benchmark, node):
...
             formatter.format_output_templates_to_be_expanded(node),
         conda:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.conda)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.conda) or "conda.yml"
         envmodules:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.envmodules)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.envmodules) or "envmodules.yml"
         container:
-            _get_environment_path(benchmark, node, SoftwareBackendEnum.apptainer)
+            _get_environment_path(benchmark, node, SoftwareBackendEnum.apptainer) or "apptainer.yml"
```
imallona, line 39: That's a quick and compact fix, thanks! But a bit obfuscated. I'd suggest adding a comment to make easier to debug for future devs.

Similarly, what if we replaced the `conda.yml` by `conda_not_provided.yml`, `envmodules.yml` by `module/not_provided/0.0.0` and `apptainer.yml` by `container_not_provided.sif` or something like that? So this is easier to read by maintainers?

bio-traven:
what is the bug here, exactly?

is it possible to add a regression test?

omni/workflow/snakemake/rules/rule_node.smk
DanInci: Thanks for the review, totally agree. Changed now and added some comments

DanInci:
@bio-traven Bug appears starting with [snakemake 8.25.2](https://github.com/snakemake/snakemake/releases/tag/v8.25.2).
If you take a look in the changes there, which is just one PR with +-150 LOC, they are not trying to base64 encode the software definitions before hashing it (in order to detect changes I assume), which was not happening before. Well, encoding a `None` throws an error, which creates us the bug.

However, having a None there is not really an issue, since the selection of software backend to run the module on is generated afterwards using a specific snakemake flag. The flag will never point to an undefined software backend, since its controlled by the same variable that is used during the definition of the rules.

bio-traven:
Thanks for the explanation! (perhaps good to include in the commit message).

I was wondering: would it work also to modify the _get function to return an empty string instead of None?

imallona (approved):
Thank you!!

DanInci:
@bio-traven yeah that would work also. Changed.
Although I lean towards the original solution as being a bit more intuitive about what's going on.

imallona:
I also prefer the original solution with explicit environment/backend strings. To me both fixes feel a bit like a smart hack to counterweight what I think is also a hack from snmk, e.g. their use of `None` to disable global variables [src](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#defining-global-container-images). Advantage of writing things like `missing/environment/caution' is that it will trigger easy to debug errors and is more readable to maintainers; returning an empty string obfuscates this and could cost debugging time.

DanInci:
@imallona reverted if we have majority

bio-traven:
good point; fwiw the question was because the boolean evaluation each time forces us to maintain changes in more places (7 at the moment, likely will increase), which can be brittle.

Additionally, comparing to None with the equality operator is generally frowned upon, from pep8:

```
Comparisons to singletons like None should always be done with is or is not, never the equality operators.

Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!
```

don't want to block this simple fix, so please consider this done. just commenting that perhaps emitting a warning is compatible with handling the default in a single place, for the next time you're working on the vicinity.

(indeed this is trying to counter a hack of using None as a special value, aka [the million dollar mistake](https://lucid.co/techblog/2015/08/31/the-worst-mistake-of-computer-science) )

DanInci:
@bio-traven Totally agree with the pep8 comment, although as long as we don't use a strongly typed language for the CLI, I really do not have a strong opinion about this, since its more like treating symptom rather than the disease in my opinion.

As for the changes in more places, totally agree with that, although I do not really expect having more than 2-3 functions for creating node rules.

PR #52, opened 2025-01-21
Aggregate generated metric results into a consolidated mappings file

DanInci:
## Ticket

No reference ticket.

## Description

The scope of this PR is to aggregate successful metric results to a so called `metrics.mappings.csv`

## Checklist:

## Remarks for the reviewer:

The approach for aggregation is done using a bottom up approach. After the execution of each module, there is a post_execution step that documents the output files in `metrics.mappings.csv`. In order for the writes to be thread safe, only append operations are allowed => multiple inserts for the same metric.

One can distinguish between the latest metric output, given the unique metric_id and latest timestamp.

TODO:
- [ ] Currently we can't distinguish between which modules are metrics and the others. Hence we map all output files from all modules. We need to add a flag in the schema for this.


imallona:
Thanks, works like a charm! I think the idea solves many problems at once.

Thinking out loud. And I'm aware this is bottom up. How do we pair this to https://github.com/omnibenchmark/omni-schema/pull/11 ? I think we will have to update this PR o the omni-schema's 11 to make them match.

The [omni-schema proposal 11]( https://github.com/omnibenchmark/omni-schema/pull/11) defines which files are metrics by listing them as inputs to a metric_collector (the `input>entries>enum`). Same mechanism as any other noninitial node, as listed below:

<details>

```yaml
[standard yml above]
metric_collectors:
  - id: biometrics
    name: "Metrics gathering and postprocessing."
    software_environment: "python"
    repository:
      url: https://github.com/omnibenchmark-example/biometrics.git
      commit: 000a001
    inputs:
      - entries:
          - metrics.mapping     ## <------------ that means these outputs are metrics 
    outputs:
      - "biometrics_report.html"
      - "biometrics.tsv"
```

</details>

This PR, though, appends an entry to `metrics.mapping.tsv` after execution of any rule, metrics or not. Hence your comment "_Currently we can't distinguish between which modules are metrics and the others. Hence we map all output files from all modules. So, to keep it bottom up, I think you're envisioning a YAML with an explicit `is_metric` flag per stage, like this :point_down: ?

<details>

```yaml
[standard yml above]
stages:
  - id: data
    modules:
      - id: D1
        name: "Dataset 1"
        software_environment: "zlib_old"
        repository:
          url: https://github.com/omnibenchmark-example/data.git
          commit: 63b7b36
    outputs:
      - id: data.counts
        path: "{input}/{stage}/{module}/{params}/{dataset}.txt.gz"
      - id: data.meta
        path: "{input}/{stage}/{module}/{params}/{dataset}.meta.json"
      - id: data.data_specific_params
        path: "{input}/{stage}/{module}/{params}/{dataset}_params.txt"

  - id: process
    modules:
      - id: P1
        software_environment: "zlib_old"
        parameters:
          - values: ["-a 0", "-b 0.1"]
          - values: ["-a 1", "-b 0.1"]
        repository:
          url: https://github.com/omnibenchmark-example/process.git
          commit: aeec1db
      - id: P2
        software_environment: "zlib_new"
        parameters:
          - values: ["-a 0", "-b 0"]
          - values: ["-a 1", "-b 0.1"]
        repository:
          url: https://github.com/omnibenchmark-example/process.git
          commit: aeec1db
    inputs:
      - entries:
          - data.counts
          - data.meta
    outputs:
      - id: process.filtered
        path: "{input}/{stage}/{module}/{params}/{dataset}.txt.gz"

  - id: some_metrics
    is_metric: true        # <------------------------------------------------------------ here
    modules:
      - id: M1
        software_environment: "zlib_old"
        repository:
          url: https://github.com/omnibenchmark-example/method.git
          commit: 8f2f835
    inputs:
      - entries:
          - data.counts
          - data.meta
          - data.data_specific_params
      - entries:
          - process.filtered
          - data.meta
          - data.data_specific_params
    outputs:
      - id: some_metrics.metrics
        path: "{input}/{stage}/{module}/{params}/{dataset}.metric.gz"
[will continue below]
```
</details>

But :point_up:  yaml still would need a dedicated `metric_collector` definition to describe which metrics are ingested, right?

<details>

```yaml
[same as above]
metric_collectors:
  - id: biometrics
    name: "Metrics gathering and postprocessing."
    software_environment: "python"
    repository:
      url: https://github.com/omnibenchmark-example/biometrics.git
      commit: 000a001
    inputs:
      - entries:
          - some_metrics.metrics  # <---------- while also having access to the metrics.mapping.tsv?
    outputs:
      - "biometrics_report.html"
      - "biometrics.tsv"

```

</details>

So, should the `metric_collector` stanza allow finetuning the input metrics via `inputs > entries> list`; or would it lack this `inputs>entries>list` part and ingest all files listed within `metrics.mapping.tsv` (being `metrics.mapping.tsv` a file listing only output files from `is_metric` stages)?

(Feature to keep in mind: we might want several metric collector modules, e.g. computing performances only, 
 or biologically-driven metrics only, or metrics run with a subset of truths etc)

DanInci:
@imallona Thanks for the detailed discussion above. I think you really highlight a lot of possible approaches downstream now with the metric collector.

We definitely need to update to [omni schema 11](https://github.com/omnibenchmark/omni-schema/pull/11), so that we have access to the metric collector. 

Agree with the 2nd point about adding a new `is_metric` flag to metric modules. This is going to allow us to capture only relevant metrics in the `metrics.mapping.tsv`. 

Regarding the 3d point, metric collector now defines the `input` field, which in my opinion, would not be really necessary, since we can make its interface generic by proxying the `metrics.mapping.tsv` directly to the collector implementation. The person that defines the collector, can filter out what he needs based on the end goal of the collector (biologically-driven metrics, subset of metrics, etc.)

At the same time, one can also figure out that `is_metric` flag can be used as a pre-filtering of metrics that you want to process. Basically if the flag is false, results are not mapped in the `metric.mapping.tsv`. This might stretch the purpose of the `is_metric` flag, but neverthless, won't hurt downstream. Considering that, can we come up with a better naming for the flag, for its dual purpose? Or leave it as it is?

Finally I envision something like this for the collector schema:
```
metric_collectors:
  - id: biometrics
    name: "Metrics gathering and postprocessing."
    software_environment: "python"
    repository:
      url: https://github.com/omnibenchmark-example/biometrics.git
      commit: 000a001
    outputs:
      - "biometrics_report.html"
      - "biometrics.tsv"
```

csoneson (approved):
I think this looks good to me. Intuitively, and in relation to the discussion above, I feel that it is probably useful to not let the `is_metric` definition do "too much" - I think it's nice to have a way to specify what is a valid metric, and then a separate way of including different sets of such valid metrics into metric collectors, it feels safer to me (sorry if I misinterpreted the comments somehow).

DanInci:
As per discussion of last week, we're gonna proceed with a top down approach, where the aggregation is done by a snakemake rule.

PR #53, opened 2025-01-29
Fix race condition when cloning github repos

DanInci:
## Ticket

No reference ticket.

## Description

When a benchmark includes the same repository as different modules, the current implementation attempts to clone it multiple times. This is not problematic in a single-threaded environment. However, when running with multiple threads, concurrent cloning of the same repository by different threads can lead to race conditions and inconsistent repository states.

## Checklist:

## Remarks for the reviewer:
- Solved using a file lock on the module directory.


imallona (approved):
Awesome, tried it and fixes the issue. Also efficient, given locks are repo-specific, thanks

PR #54, opened 2025-02-03
Add benchmark directives to rules (without collection)

imallona:
This is a slim variant of #45

## Ticket

No ticket.

## Description

Collect benchmarks per rule/node, i.e.

```
$ cat ./out/data/clustbench/param_0/clustering/fastcluster/param_1/metrics/partition_metrics/param_7/clustbench_performance.txt
s       h:m:s   max_rss max_vms max_uss max_pss io_in   io_out  mean_load       cpu_time
3.6763  0:00:03 70.93   79.91   59.22   64.69   24.13   0.01    0.00    1.04
```

## Checklist:

## Remarks for the reviewer:

Collection/aggregation is still pending, this generates one file per rule (run).

### DIFF
--- a/omni/workflow/snakemake/format/formatter.py
+++ b/omni/workflow/snakemake/format/formatter.py
@@ -2,6 +2,7 @@ import re
 from itertools import takewhile
 from pathlib import Path
 from typing import List, Set, Tuple, Union, NamedTuple
+import os.path as op
 
 from omni.benchmark import BenchmarkNode, Benchmark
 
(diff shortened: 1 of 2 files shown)


imallona:
Re: the Mac tests. I would remove support to windows and Mac. 

To me running bionfo benchmarks on a Mac machine is not an usecase (who does this?), and the whole arm, easybuild, envmodules etc feels like making our life unnecessarily complicated.

DanInci:
@imallona I agree that mac support complicates things, especially when using software backends. Windows support was already dropped right?
But I do believe people are going to test omnibenchmark out on mac before deciding to run a proper benchmark on HPC.

Maybe we can drop software support from mac machines.

DanInci (changes requested):
It's missing the benchmark directive from standalone node workfows:
`create_standalone_node_rule`

These are already refactored and included in my other PR: https://github.com/omnibenchmark/omnibenchmark/pull/52

I can build on top of that one and include this + the new metrics mapping implementation that we talked about on the last meeting.

EDIT: Now included in PR https://github.com/omnibenchmark/omnibenchmark/pull/53

imallona:
Thanks, closing so #56 does this instead

PR #57, opened 2025-02-06
Failed attempt to add extra flag (akin to --keep-going)

imallona:
This is an "understanding the codebase" question. I'm trying to add a `--keep-going` snakemake flag when the CLI has it (well, `--continue-on-error`). What I am missing to be able to make it run? The snakemake call does not recognize `keep_going` as being `True` even after passing the CLI command.

### DIFF
--- a/omni/cli/run.py
+++ b/omni/cli/run.py
@@ -149,6 +149,13 @@ def run_benchmark(
     default=False,
 )
 @click.option(
+    "-k",
+    "--continue-on-error",
+    help="Go on with independent jobs if a job fails (--keep-going in snakemake).",
+    is_flag=True,
+    default=False,
+)
+@click.option(
     "--keep-module-logs/--no-keep-module-logs",
     default=False,
     help="Keep module-specific log files after execution.",
--- a/omni/cli/run.py
+++ b/omni/cli/run.py
@@ -44,6 +44,7 @@ def run(ctx):
     default=False,
 )
 @click.option("-d", "--dry", help="Dry run.", is_flag=True, default=False)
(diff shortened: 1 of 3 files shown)


omni/cli/run.py
```diff
@@ -68,6 +69,13 @@ def run_benchmark(ctx, benchmark, threads, update, dry, local, keep_module_logs)
     benchmark = validate_benchmark(benchmark)
     workflow: WorkflowEngine = SnakemakeEngine()
 
+    ## it is as --continue_on_error originally
+    if continue_on_error:
+        keepgoing_prompt = click.confirm(
+            "Are you sure you want to run the full benchmark even if some jobs fail?", abort=True
+        )
+        if not keepgoing_prompt:
+            raise click.Abort()
```
imallona, line 78: This was to make sure the flag works as expected; it is recognized

omni/workflow/snakemake/snakemake.py
```diff
@@ -325,6 +332,9 @@ def _prepare_argv(
         if dryrun:
             argv.append("--dryrun")
 
+        if continue_on_error:
+            argv.append("--keep-going")
+
         if (
```
imallona, line 338: This is not happening, snakemake does not receive the arg `--keep-going`

DanInci:
@imallona Problem is that the order of the passing arguments to _prepare_argv is incorrect. `keep_module_logs` is first, followed by `continue_on_error` second. But they were passed the other way around. Fixed it by using named arguments to the call.

imallona:
Thank you @DanInci!

PR #58, opened 2025-02-26
Release v0.2.0

imallona:
Added metric collectors, archival, `--continue-on-error`, and fixes (repo cloning)

## Checklist:

# Extra checklist
- [ ] I have updated the omnibenchmark.org tutorial / others accordingly
- [ ] I will update the omnibenchmark.org CLI reference accordingly 
  - [ ] I am aware of and have proactively fixed this https://gitlab.renkulab.io/omnibenchmark/omni_site/-/merge_requests/20 before merging


pyproject.toml
```diff
@@ -40,7 +40,7 @@ isort = "^5.13.2"
 matplotlib = "3.8.0"
 networkx = "^3.3"
 
-omni-schema = {git = "https://github.com/omnibenchmark/omni-schema.git", rev = "version/0.0.4"}
+omni-schema = {git = "https://github.com/omnibenchmark/omni-schema.git", rev = "94c312957f96d218369e5e1bcf7abaf976ed0fbc"}
```
imallona, line 43: we'd benefit from a new omni-schema release. @btraven00 @bio-traven would you like to discuss (so you learn the flow here)?

imallona: btw, which user are you going to contribute with, Ben? so I remove the other one from the org for the sake of clarity?

imallona:
We haven't updated the omnibenchmark.org tutorial/docs to discuss metric collectors and archival, any volunteer?

pyproject.toml
btraven00: happy to discuss, both current state and desired. I was wondering about the benefits of having the schema as a separate repo with its own release chores, is this documented somewhere? happy to learn/help with the release process

btraven00:
I don't mind updating docs, it might help me understanding the current design/spec.

pyproject.toml
imallona: Thanks. Long story short, https://github.com/omnibenchmark/omnibenchmark?tab=readme-ov-file#developer-notes

DanInci: @btraven00 Right now, the benefits may not be immediately obvious, but I believe having the data backbone of the omni project centralized would be useful in the long run. For example, if a future project aims to implement the CLI in Rust, it could use the same backbone instead of reimplementing it (leading to forks of the data model). Similarly, other projects could be built on top of it, making the ecosystem more modular and reusable.

DanInci (dismissed):

DanInci (commented):
We should publish a new version of the omnibenchmark, so as not to depend on a snapshot dependency when creating a new release

imallona:
Going back to draft state because rc4 should incorporate #80 and #79, at least.

btraven00:
I think we should tag + upload 0.1.0 from main, at this point, and reschedule this release as 0.2.0 (enough features/changes to guarantee minor ver bump)

PR #59, opened 2025-02-26
cli for repo creation

retogerber:
## Description

Draft of how to help setup repos of modules. Adds two new cli commands:
```
ob info repo list-missing -b tests/data/Clustering.yaml
```
which lists node names where the remote repository doesn't exist (repo url + commit, currently only checks github/gitlab).
and
```
ob info repo create-missing -b tests/data/Clustering.yaml -m data-fake-default -o tmp
```
which creates two files, main.(py/R) and config.cfg, which should help to quickly setup methods. The files are written to directory "tmp".

## Checklist:

### DIFF
--- a/data/templates/main.py
+++ b/data/templates/main.py
@@ -1,0 +1,35 @@
+import argparse
+import sys
+import os
+
+
+def main():
+    parser = argparse.ArgumentParser(description="")
+
+    # Inputs
+    INPUTS_ARGPARSE
+
+    # Parameters
+    PARAMETERS_ARGPARSE
+
+    # Outputs
+    parser.add_argument(
+        "--output_dir", type=str, help="output directory to store data files."
+    )
+    parser.add_argument("--name", type=str, help="name of the dataset", default="")
+
+    try:
+        args = parser.parse_args()
+    except:
+        parser.print_help()
+        sys.exit(0)
+
+    VARIABLES_INPUTS
+
+    VARIABLES_PARAMETERS
+
+    VARIABLES_OUTPUTS
+
+
+if __name__ == "__main__":
+    main()
--- a/omni/benchmark/template_repo.py
+++ b/omni/benchmark/template_repo.py
@@ -1,0 +1,196 @@
+from pathlib import Path
+from typing import List
+import os
+
+from omni.benchmark import Benchmark
+from omni.io.code import get_git_archive
+
+# parse benchmark
+
+# get all repositories
+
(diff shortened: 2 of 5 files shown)


omni/benchmark/template_repo.py
```diff
@@ -0,0 +1,201 @@
...
+from omni.io.code import get_git_archive, check_remote_repo_existance
+
+# parse benchmark
+
+# get all repositories
+
+# check if repository (and commit hash) exists (e.g. is reachable), and if yes if actually is omnibenchmark repo, i.e. contains config.cfg file and the specified file in the config.cfg file
+
+# if not, create a new repository:
+# - figure out which language: python, R (by description/name of software env used?)
```
imallona, line 15: I guess this means check the softenv spec -> guess the interpreter. But why not the way around? Meaning, if a thing is clearly R then we could suggest a base R apptainer sif/conda yaml/easyconfig? Because, in the past [we discussed the method developer must provide an env spec to test their thing](https://github.com/omnibenchmark/internal_docs/blob/master/architecture/method_contributor_design.md). Up to the benchmarker to reuse this spec within the benchmark YAML or not.

data/templates/main.py
```diff
@@ -0,0 +1,35 @@
...
+        args = parser.parse_args()
+    except:
+        parser.print_help()
+        sys.exit(0)
+
+    VARIABLES_INPUTS
+
+    VARIABLES_PARAMETERS
+
+    VARIABLES_OUTPUTS
```
btraven00, line 31: hmm what is this supposed to be? 

if these (and the blocks above) are placeholders, it would be good to either comment the block with some explanation (boilerplate to be filled by the user), or to enact some mechanism that avoids the boilerplate from running if the basic required implementations are not provided

btraven00:
First, I think I like the intention here. Two quick high level comments:

- It would be easier for a relative outsider to review if there'd be a high-level description of the *intent* of the PR. True that there's some behavior-level explanation in the ticket, but what I'm missing is a bit more about what operations is the code expected to perform (i.e., what problem is the code solving).
- I mention this because in a quick read it seems to me that this PR is solving more than the stated problem (i.e., it implicitly enters the problem space of defining per-implementation entrypoints etc). 

Sure, both problems (scaffolding and entrypoints) go hand in hand, but perhaps we could benefit of discussing those problems at different levels of abstraction :)

retogerber:
Thanks for your comments! I agree that this should probably be drafted as an abstraction first before implementation. I opened #67 to track and discuss that there.

Issue #70, opened 2025-04-02
fix existing type annotations


btraven00:
As I start working on different areas of the existing codebase, I noticed that some areas already have type annotations.

This is very nice. But at some point we need  to start enforcing type checking.

Low-prio for me, but as I think it can help with documentation and reasoning about critical paths, I think we should at least begin to be aware of when the included type annotations are inconsistent.

PR #71, opened 2025-04-02
tests: add basic pyrightconfig

btraven00:
## Descriiption 

Add a preliminary pyrightconfig file for starting fixing type
annotations.

As I start touching different areas of the codebase, I plan to slowly
fix all the ~200 errors that the current tree (dev branch) currently
throws. I suspect that some of these are minor inconsistencies, but some
actual bugs might be laying among them.

* Related: #70


btraven00:
rebased & merged

Issue #65, opened 2025-03-18
implement a timeout policy for long-running tasks


btraven00:
We would like to specify a "sane" default for a mask time limit allowed for each task execution. This might be dynamically set (i.e., expressed as a percentile of the completed tasks), but that is probably better left to a later iteration.

For starters, I think we could assume that a max task duration is set on the benchmark specification, in seconds. Thoughts?

At the very least while testing and debugging runs, this would allow for faster iteration cycles. This could imply that we might want to set a hard limit configured at the level of the cli execution (see #31), instead of making it a property of a particular benchmark. But I think a per-benchmark default (perhaps overridable by a flag) is not unreasonable.

imallona:
If thinking of reusing snakemake `resources` handling (not only for wallclock) please be aware of [local vs remote]( https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources-and-remote-execution). Perhaps easier if we wrapped the executor with some timeout, profiler etc ourselves.

btraven00:
The issue is that resources is just a convenience annotation that cannot be enforced locally, as you said. However, it's trivial to enforce it since we're doing a (weirdly, double) spawning of the entrypoints.

@imallona what would be your preference for exposing this in ob? I'm inclined to do a single `--timeout` directive in the cli, and rather not touch the spec right now (for reasons I've expressed elsewhere).

I think this implementation is ~good enough for allowing us to limit the long tail of run times, in the future I can see that different tasks (rules) can get different policies in the yaml.

One tangential concern is: how should we handle timeouts in terms of not-successful-runs? Right now it's a failure as the others (and I'm fine leaving it as is). But, since there's no output I assume the metric ollection is not going to happen as long as there's at least one timeout. Should we touch an empty output file? Should we differentiate timeouts from other errors? (e.g., in the logs)

PR #79, opened 2025-04-07
feat: pass an optional timeout argument to abort long-running local tasks

btraven00:
## Ticket

https://github.com/omnibenchmark/omnibenchmark/issues/65

## Description

This feature adds a resource annotation that is enforced via the local execution, by passing the optional timeout parameter (in seconds) to the subprocess invocation. It is exposed via a single CLI flag.

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,10 +162,18 @@ cython_debug/
 # tildes
 *~
 
+# pixi environments (not using it officially yet)
+.pixi
+pixi.lock
+pixi.toml
+
+# snakemake stuff
+Snakefile
+.snakemake/
+
 # Custom omni-benchmark specific folders
 out/
 log/
-.snakemake/
 output_dag.png
 Snakefile
 
--- a/omnibenchmark/benchmark/constants.py
+++ b/omnibenchmark/benchmark/constants.py
@@ -1,0 +1,10 @@
+from typing import Final
+
+from humanfriendly import parse_timespan
+
+# DEFAULT_TIMEOUT_HUMAN is the human representation of the max value we'll try to enforce
+# as a timeout for the spawned task.
+# This is currently only enforced on local environments; when supporting slurm we can pass
+# it as runtime resource.
+DEFAULT_TIMEOUT_HUMAN: Final[str] = "4h"
+DEFAULT_TIMEOUT_SECONDS: Final[int] = int(parse_timespan(DEFAULT_TIMEOUT_HUMAN))
--- a/omnibenchmark/cli/benchmark.py
+++ b/omnibenchmark/cli/benchmark.py
@@ -154,12 +154,13 @@ def list_versions(ctx, benchmark):
     envvar="OB_BENCHMARK",
 )
 @click.pass_context
-def plot(ctx, benchmark: str):
+def computational_graph(ctx, benchmark: str):
     """Export computational graph to dot format."""
 
     b = validate_benchmark(benchmark, echo=False)
-    dot = b.export_to_dot()
-    click.echo(dot.to_string())
+    if b is not None:
+        dot = b.export_to_dot()
+        click.echo(dot.to_string())
 
 
 @info.command("topology")
@@ -176,5 +177,6 @@ def plot_topology(ctx, benchmark: str):
     """Export benchmark topology to mermaid diagram format."""
 
     b = validate_benchmark(benchmark, echo=False)
-    mermaid = b.export_to_mermaid()
-    click.echo(mermaid)
+    if b is not None:
+        mermaid = b.export_to_mermaid()
+        click.echo(mermaid)
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -6,7 +6,9 @@ from itertools import chain
 from pathlib import Path
 
 import click
+import humanfriendly
 
+from omnibenchmark.benchmark.constants import DEFAULT_TIMEOUT_HUMAN
 from omnibenchmark.cli.utils.logging import logger
 from omnibenchmark.cli.utils.validation import validate_benchmark
 from omnibenchmark.io.utils import get_storage_from_benchmark
@@ -37,7 +39,7 @@ def run(ctx):
 @click.option(
     "-c",
     "--cores",
-    help="Use at most N CPU cores in parallel. Default is all available CPU cores",
+    help="Use at most N CPU cores in parallel. Default is 1.",
     type=int,
     default=1,
 )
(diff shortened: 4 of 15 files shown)


omni/workflow/snakemake/rules/rule_node.smk
```diff
@@ -49,7 +65,9 @@ def _create_initial_node(benchmark, node, config):
             parameters = node.get_parameters(),
             dataset = module_id,
             keep_module_logs=config['keep_module_logs']
-        script: os.path.join(os.path.dirname(os.path.realpath(scripts.__file__)), 'run_module.py')
+        resources:
+            timeout=timeout
+        script: SCRIPT_PATH
```
DanInci, line 70: Why not use the import path, instead of hardcoding the `SCRIPT_PATH` ?

omni/workflow/snakemake/snakemake.py
```diff
@@ -212,6 +217,7 @@ def serialize_node_workflow(
         node: BenchmarkNode,
         output_dir: Path = Path(os.getcwd()),
         write_to_disk=True,
+        timeout: int = -1,
```
DanInci, line 220: Why default to -1 ?

tests/cli/test_run_benchmark_timeout.py
```diff
@@ -0,0 +1,93 @@
...
+from importlib.resources import files
+
+
+def get_script_path():
+    # Get the package-relative path
+    script_path = files("omni.workflow.snakemake.scripts").joinpath("run_module.py")
+    return str(script_path)
+
+
+SCRIPT_PATH = get_script_path()
```
DanInci, line 30: Can we set this as a fixture in `conftest.py` ?

tests/data/mock_benchmark.yaml
```diff
@@ -1,10 +1,7 @@
 id: mock_benchmark
-description: simple benchmark, user for fast testing
+description: simple benchmark, used for fast testing
 version: 1.0
 benchmarker: "John Doe at Robinsons lab, john.doe@uzh.ch"
-storage: https://storage.github.com/
-storage_api: S3
-storage_bucket_name: mock_benchmark
 benchmark_yaml_spec: 0.01
```
DanInci, line 5: Existing tests are failing now because the benchmark schema definition requires these fields. Either we keep them here, or remove them from the schema definition first.

DanInci (changes requested):

tests/data/mock_benchmark.yaml
btraven00: https://github.com/omnibenchmark/omni-schema/pull/16 I guess we can tag 0.0.6

omni/workflow/snakemake/snakemake.py
btraven00: Good question. I was passing -1 to mark an invalid value, and then at some point I added the constants default (4 hours). Do you think we should put the default here too?

tests/cli/test_run_benchmark_timeout.py
btraven00: right now only this test seems to need it, so I'd keep it local. but yea definitely keep an eye if the problem surfaces somewhere else and then factor it out?

tests/data/mock_benchmark.yaml
DanInci: Indeed, that would be the diea

omni/workflow/snakemake/rules/rule_node.smk
DanInci: No no, I mean what happened to:

```python
from omni.workflow.snakemake import scripts

os.path.join(os.path.dirname(os.path.realpath(scripts.__file__)), 'run_module.py')
```

Woudn't it be more safe to use `scripts.__file__` instead of hardcoding "omni.workflow.snakemake.scripts"?

omni/cli/run.py
```diff
@@ -63,9 +64,23 @@ def run(ctx):
     default=False,
     help="Keep module-specific log files after execution.",
 )
+@click.option(
+    "--timeout",
+    type=int,
+    default=DEFAULT_TIMEOUT_SECONDS,
```
imallona, line 70: same, "minutes" feels friendlier to me

omni/workflow/snakemake/rules/rule_node.smk
```diff
@@ -49,7 +65,9 @@ def _create_initial_node(benchmark, node, config):
             parameters = node.get_parameters(),
             dataset = module_id,
             keep_module_logs=config['keep_module_logs']
-        script: os.path.join(os.path.dirname(os.path.realpath(scripts.__file__)), 'run_module.py')
+        resources:
+            timeout=timeout
```
imallona, line 69: is `timeout` a valid snakemake resource param? `runtime` is but `timeout` isn't afaik

imallona: perhaps helpful from snmk's codebase: [job sorting/scheduling/grouping docs](https://github.com/snakemake/snakemake/blob/6800a2baf5500a009c01c9d3585d607fd276e8c8/src/snakemake/resources.py#L158) taking into account resources runtime

omni/workflow/snakemake/rules/rule_node.smk
btraven00: Oh I see. Thanks for clarifying!

So, the problem here is that we should not rely on `__file__` (it's ~fine for informal usage).

From https://docs.python.org/3/reference/datamodel.html: 
```
module.__file__
module.__cached__

    __file__ and __cached__ are both optional attributes that may or may not be set. Both attributes should be a [str](https://docs.python.org/3/library/stdtypes.html#str) when they are available.
```

omni/benchmark/constants.py
```diff
@@ -0,0 +1,5 @@
+from typing import Final
+
+# DEFAULT_TIMEOUT_SECONDS is the max value we'll try to enforce as a timeout for the spawned task.
+# This is currently only enforced on local environments (but we can pass the same value to job queues)
+DEFAULT_TIMEOUT_SECONDS: Final[int] = 4 * 60 * 60  # Default is 4 hours
```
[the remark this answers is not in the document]
btraven00: I have one objection though, and it's that at least for testing I'd like to retain seconds. Using fractions of a minute will look weird. So, at least for internal representation, I'd like to retain ints and seconds. See my comment below for the flag.

omni/cli/run.py
btraven00: Are you ok with parsing human-friendly quantities instead?

Example:

```
--timeout 10m
--timeout 2h
--timeout 1d
--timeout 30s
```

omni/workflow/snakemake/snakemake.py
btraven00: I'm pretty sure it is working - see the test. Also see my comment in https://github.com/omnibenchmark/omnibenchmark/pull/79#discussion_r2035083776

btraven00: That said, I'm gonna fix the -1 to a more sensible default (whatever subprocess timeout takes a default)

omni/cli/run.py
imallona: using human-friendly quantities sounds awesome to me, thanks

DanInci (approved):
LGTM, once the tests & code style checks pass.

imallona (approved):

btraven00:
need to be rebased on top of current dev

Issue #73, opened 2025-04-03
storage_api should be optional


btraven00:
I'd rather not have to specify a s3 storage API for a test benchmark.

retogerber:
Agree.

btraven00:
this has been solved

PR #80, opened 2025-04-09
chore: cleanup dependencies

btraven00:
## Ticket

Related to https://github.com/omnibenchmark/omnibenchmark/issues/73

## Description

Move the three s3-related dependencies to an optional dependency block.

While doing this, I updated the pyproject.toml to be compatible with the latest poetry version. Also did a bit of housekeeping, since some of the main dependencies were just dev dependencies, and some others (typing-extensions) are not really needed if we're supporting python 3.12+

Linter has been changed from black to ruff (faster and more complete). This includes the pre-commit hook.

Additionally, adapted some of the tests so that they pass the checks. Notably, there's one conda+snakedeploy test that was not reaching completion. This, together with some flakyness in the easybuild tests, make me think that we could benefit of increasing the latency wait in snakemake, since it seems some of the tests are finishing before we're able to collect all of the output. Postponing that to a different PR.

## Checklist:


btraven00:
forcing cache cleanup in github runner

imallona (approved):

omni/software/conda_backend.py
```diff
@@ -13,5 +12,5 @@ def pin_conda_envs(benchmark_yaml: Path) -> None:
     Pins a conda env file using snakedeploy
     """
     snakeconda.pin_conda_envs(
-        conda_env_paths=[benchmark_yaml], conda_frontend="mamba", warn_on_error=False
+        conda_env_paths=[benchmark_yaml], conda_frontend="mamba", warn_on_error=True
     )
```
DanInci, line 16: Should logging.level be checked here, and depending on it set the flag?

tests/software/conda_tests.py
```diff
@@ -32,8 +30,8 @@ def test_conda():
 
 
 def test_omni_python_import():
-    import omni
+    pass
 
 
 def test_omni_easybuild_import():
-    from omni.software import easybuild_backend as easy
+    pass
```
DanInci, line 37: Are these tests relevant anymore? I assume the pre-commit hook removed the unused imports.

tests/software/envmodules_tests.py
```diff
@@ -11,23 +11,20 @@
...
 
 
 def test_omni_python_import():
-    import omni
+    pass
 
 
 def test_omni_easybuild_import():
-    from omni.software import easybuild_backend as easy
+    pass
```
DanInci, line 27: Same here 2. Are these tests relevant anymore? I assume the pre-commit hook removed the unused imports.

tests/software/singularity_tests.py
```diff
@@ -63,11 +61,11 @@ def test_bash_cmd():
 
 
 def test_omni_python_import():
-    import omni
+    pass
 
 
 def test_omni_easybuild_import():
-    from omni.software import easybuild_backend as easy
+    pass
```
DanInci, line 68: Same here 3. Are these tests relevant anymore? I assume the pre-commit hook removed the unused imports.

DanInci (changes requested):

pyproject.toml
```diff
@@ -1,80 +1,102 @@
...
+    # DAG & Visualization
+    "networkx>=3.3", # this might be overkill for the DAG
+    "matplotlib==3.8.0",
+    "pydot>=3.0.2",
+
+    # Cloning modules
+    "GitPython>=3.1.0",
+
+    # Workflow
+    "snakemake==8.25.1", # TODO: try to unpin
```
imallona, line 43: re: the unpinning while also making tests faster/easier. I think currently snmk is the major dependency and the most likely to break things. Worth updating the github actions to try several snmk versions (latest, perhaps 8.21.1 for v8, and  9.1.7 or whatever relevant for v9)? I think setting up the grid of snmk versions would save us time and pain when some capabilities are moved to the plugin universe. Hence having to be proactive and able to tune `snakemake` calls to define the relevant plugin(s) and its arguments.

omni/software/conda_backend.py
btraven00: I disabled it because tests were failing based on what I inferred was this warning.

I need to check if the silencing was critical or not.

Regardless, I don't think it's related to our logging level; if set to true, [this will raise an exception](https://github.com/snakemake/snakedeploy/blob/main/snakedeploy/conda.py#L162) that I'm not sure it's been watched elsewhere

tests/software/conda_tests.py
btraven00: Good question. Yea I think you're right, this is the linter removing unused.

How do you feel about them? From my point of view, these tests are just testing that the test environment correctly installed the needed dependencies. I'm not sure if I see how they belong to conda_tests.

But if you want to retain them, the way to go might be to assert a property of  the modules (not None, for instance).

pyproject.toml
btraven00: Agree. As I would do it, I would keep the most minimal pinning possible here on the deps; and then keep a [lockfile (or even a pypi requirements)](https://docs.astral.sh/uv/concepts/projects/sync/) with the known working versions; plus possibly a specific conda env yaml as you're currently doing. 

Since it's going on a slightly tangent, I'd handle the snakemake testing as a sep issue at this point.

tests/software/conda_tests.py
DanInci: I agree with the refactoring. Maybe want to double check with @imallona about the reason why they are in conda tests.

DanInci (approved):

Issue #90, opened 2025-04-29
add git bundles for more self-contained testing


btraven00:
instead of relying on cloning repos over the network, we can embed simple git repo bundles with all the test files we need for the simple tests.

for sure we want to exercise cloning from git remotes too, but that's left as an integration test.

btraven00:
mechanism is in place and documented. it's up to test implementors to make use of it.

PR #91, opened 2025-04-29
tests: first implementation of repo bundles for testing

btraven00:
## Ticket

https://github.com/omnibenchmark/omnibenchmark/issues/90

## Description

## Checklist:

## Remarks for the reviewer:

Just a first PoC. We will need to think if the scope is enough or not, and overall where do we want to checkout the repo to (it might be that we need to templatize the ob yaml. for now I think we can live with having all the repos cloned at tests/data/repos once per session, but this has the limitation that we can use only one commit per repo per yaml). still, worth trying first I tink.


imallona:
Thanks! "_but this has the limitation that we can use only one commit per repo per yaml_": why is this a limitation and not a feature? I think we could/should clone and then checkout the relevant commit. 

(Independently, it's a pity we have to clone without specifying maxdepth or a branch in order to be sure we can find the requested commit...)

btraven00:
I've used this work in #128 , closing this PR
at the end, there is not such limitation. Right now you can have as many bundles as you want. Each operation will clone the passed commit. 

I think this is good enough for the kind of polite and lightweight repos that we want to use as tests.

I'm leaving the migration of existing repos to use bundles as an exercise for future versions of myself.

PR #93, opened 2025-04-29
tests: simplify benchmark used in tests

DanInci:
## Ticket

No ticket reference.

## Description

Refactor parameter hashing tests in order to be simpler and take less time.

## Checklist:


btraven00:
also, can you merge dev on top of that? I think I don't see a clear diff against dev..

tests/data/Benchmark_003_trimmed.yaml
```diff
@@ -0,0 +1,139 @@
+id: Benchmark_001
```
btraven00, line 1: needs update no?

tests/data/Benchmark_003_trimmed.yaml
```diff
@@ -0,0 +1,139 @@
...
+        repository:
+          url: https://github.com/omnibenchmark-example/process.git
+          commit: 3b17081
+      - id: P2
+        software_environment: "zlib_new"
+        parameters:
+          - values: ["-a 0", "-b 0"]
+          - values: ["-a 2", "-b 0.5"]
+        repository:
+          url: https://github.com/omnibenchmark-example/process.git
```
btraven00, line 74: this is using R. see https://github.com/btraven00/omnibenchmark-testing-process for a pythonic version

tests/data/Benchmark_003.yaml
```diff
@@ -1,51 +1,49 @@
-id: benchmark003
+id: Benchmark_001
```
btraven00, line 1: needs change?

tests/data/Benchmark_003_trimmed.yaml
```diff
@@ -0,0 +1,139 @@
+id: Benchmark_001
+description: simple benchmark, somewhat explicit, simple params
```
btraven00, line 2: maybe add in the description what differences are there with the base yaml and what tests are using this.

tests/data/Benchmark_003_trimmed.yaml
```diff
@@ -0,0 +1,139 @@
...
+        path: "{input}/{stage}/{module}/{params}/{dataset}.meta.json"
+      - id: data.data_specific_params
+        path: "{input}/{stage}/{module}/{params}/{dataset}_params.txt"
+
+  - id: process
+    modules:
+      - id: P1
+        software_environment: "zlib_old"
+        parameters:
+          - values: ["-a 0", "-b 0"]
```
btraven00, line 63: watch out, I think the single dash needs to be converted to double (depends on the version of the args serializer in use). In the python version of process I've used --

btraven00 (changes requested):

tests/data/Benchmark_003_trimmed.yaml
DanInci: The latest commit is not using R: https://github.com/omnibenchmark-example/process

tests/data/Benchmark_003_trimmed.yaml
DanInci: It works with both. I adapted the parameter parser.

tests/data/Benchmark_003_trimmed.yaml
btraven00: ah, so this is the one you changed, I see! I was misled because I had forked the repo for a different test.

btraven00 (approved):
:whale:

Issue #88, opened 2025-04-29
Split tests into fail-early pipeline


btraven00:
We can get more fancy, but this should work for starters:

1. Run quick tests (smoke and unit tests)
2. Run integration tests (simple benchmarks, depends on containers, like remote storage etc)
3. Run build (i.e. generate wheel, perhaps expose it as artifact)
4. Run software tests (reuse the wheel from previous step). This is better done from within a docker image that has everything installed.

To catch problems with updates to dependencies in the chain (conda, etc) we can setup a separate test matrix (nightly, or in a separate repo).

For incremental changes, 1 is possibly the lower-hanging fruit (i.e. use `pytest -m short` vs the rest).

btraven00:
Closing this since we merged #94. It can be further improved, but the pipeline is split.

PR #94, opened 2025-04-29
tests: split pipeline

btraven00:
## Ticket

Related to https://github.com/omnibenchmark/omnibenchmark/issues/88

## Description

First split of the test pipeline. More work needs to be done for better separation of concerns, and using a more streamlined image.

## Checklist:

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -1,7 +1,11 @@
 name: Code Style Check
 
 on:
-  pull_request:
+  # Removed pull_request trigger to avoid duplication with pipeline.yml
+  workflow_dispatch: # Allow manual trigger if needed
+  push:
+    branches:
+      - main # Still run on main branch pushes
 
 jobs:
   code-style:
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -15,18 +15,10 @@ jobs:
           python-version: "3.12"
 
       - name: Install Linters
-        run: pip install black ruff
+        run: pip install ruff
 
-      # For the time being, let's keep the black check as the
-      # only mandatory step.
-      - name: Check coding style
-        run: black --check .
-
-      # Optional Ruff check (allowed to fail)
-      # Ruff if faster, but it's throwing a wider variety of errors that still need
-      # work to be fixed.
-      - name: Lint with Ruff (optional)
-        continue-on-error: true # 👈 This makes the job pass even if Ruff fails
+      # Ruff check
+      - name: Lint with Ruff
+        continue-on-error: false
         run: |
           ruff check .
-          echo "Note: Ruff failures are non-blocking for now"
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Run longer tests (integration)
         run: |
-          poetry run pytest -m "not short" --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml \
+          poetry run pytest -m "not short and not conda" --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml \
               tests/benchmark tests/cli tests/workflow tests/io
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
 
@@ -117,7 +117,9 @@ jobs:
         os: [ubuntu-22.04]
         test_group: [1, 2]
         python-version: ["3.12"]
-      fail-fast: false
+      fail-fast: true
+
+    # TODO: run on particular container with conda etc
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -84,6 +84,22 @@ jobs:
           poetry run pytest -m short
         if: matrix.os == 'ubuntu-latest'
 
+      # Build wheel after short tests pass
+      - name: Build wheel
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
+        run: |
+          pip install build
+          python -m build .
+
+      # Upload the wheel as an artifact
+      - name: Upload wheel artifact
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
+        uses: actions/upload-artifact@v3
+        with:
+          name: omnibenchmark-wheel
+          path: dist/*.whl
+          retention-days: 7
+
       - name: Run longer tests (integration)
         run: |
           poetry run pytest -m "not short and not conda" --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml \
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -65,7 +65,7 @@ jobs:
 
       - name: Load cached venv
         id: cached-dependencies
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             .venv
@@ -94,7 +94,7 @@ jobs:
       # Upload the wheel as an artifact
       - name: Upload wheel artifact
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: omnibenchmark-wheel
           path: dist/*.whl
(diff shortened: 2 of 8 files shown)


tests/software/test_run_with_software.py
```diff
@@ -245,6 +248,7 @@ def test_multi_run_workflow_with_parameters():
 
 
 @skip_if_no_conda
+@pytest.mark.conda
 def test_multi_run_workflow_with_parameter_removal():
     benchmark_file = Path("..") / "data" / "Clustering.yaml"
     benchmark_file_path = Path(__file__).parent / benchmark_file
```
DanInci, line 254: Expect conflicts here, since these tests were moved to `test_run_workflow.py` and no longer using backends.

.github/workflows/code-style.yml
```diff
@@ -1,7 +1,11 @@
 name: Code Style Check
 
 on:
-  pull_request:
+  # Removed pull_request trigger to avoid duplication with pipeline.yml
+  workflow_dispatch: # Allow manual trigger if needed
+  push:
+    branches:
+      - main # Still run on main branch pushes
```
DanInci, line 8: I don't think the code style check should run on main / dev branches.
If we stick to pull request + protection on main branches, then code that doesnt pass the code style check won't be integrated.

Here I'll leave the triggers on `pull_request` and `workflow_dispatch`

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,271 @@
...
+        run: poetry check
+
+      - name: Load cached venv
+        id: cached-dependencies
+        uses: actions/cache@v3
+        with:
+          path: |
+            .venv
+            ${{ env.R_LIBS_USER }}
+          key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
```
DanInci, line 73: ${{ env.R_LIBS_USER }} is not needed anymore, since there's no R leftover.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,271 @@
...
+      - name: Run longer tests (integration)
+        run: |
+          poetry run pytest -m "not short and not conda" --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml \
+              tests/benchmark tests/cli tests/workflow tests/io
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
+
+      - name: Run tests (Windows)
+        run: |
+          poetry run pytest --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml
+        if: matrix.os == 'windows-latest'
```
DanInci, line 97: Windows tests are no longer supported and `windows-latest` is not in the os matrix.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,271 @@
...
+      - name: Run short tests
+        run: |
+          poetry run pytest -m short
+        if: matrix.os == 'ubuntu-latest'
+
+      - name: Run longer tests (integration)
+        run: |
+          poetry run pytest -m "not short and not conda" --cov=omnibenchmark --cov-report=xml:./reports/coverage.xml \
+              tests/benchmark tests/cli tests/workflow tests/io
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
```
DanInci, line 92: I wonder if longer tests (integration) should be a separate test in the pipeline, between short tests and software tests?

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,271 @@
+name: CI Pipeline
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
```
DanInci, line 8: Now I wonder why decide on keeping both this pipeline.yml and the previous existing action files?

The only difference I expect is the triggers. Basically here is the definition for the pull request trigger, and in the others for the manual and main branches trigger.

Can we just add the same triggers in here?

.github/workflows/code-style.yml
DanInci: Also I believe we want to run it also on `dev`

tests/workflow/test_run_workflow.py
```diff
@@ -47,6 +49,11 @@ def test_run_workflow_001(snakemake_env, tmp_path):
...
 
 
+# FIXME: this test needs conda right now
+# but in reality this behaves like a software backend test. I think the proper way to test this is that
+# we proceed with the workflow execution without raising any error, but that needs
+# to be implemented in the future.
+@pytest.mark.conda
 def test_run_workflow_backends_missing(snakemake_env, tmp_path):
     # Use current directory if specified, otherwise use tmp_path
     # tmp_path is already provided by pytest fixture if not using current dir
```
DanInci, line 59: I forgot what was the purpose of this test? It seems redundant now, since it just tests a normal workflow run with conda.

DanInci (changes requested):

.github/workflows/code-style.yml
btraven00: So, my strategy for now is to duplicate the current test suite and the new pipeline, and compare for a bit.
You're right that normal PR process should be enough.

otoh, I think it's sane to run the same suite for every merge - nothing should happen, but at least we should get a confirmation that tests are still green after a while.

I'm not comfortable with the code duplication, but I think I prefer to leave what we know as the green state for a bit, so that we can compare side by side. And then probably delete the legacy tests once we're comfortable with the pipeline.

Couple notes:

- Yes, dev should be tested too.
- Right now they're basically the same, but I plan on iterate on the split pipeline.
- Ideally parallelizing the software environments too (conda vs envmodules vs apptainer etc).
- FWIW, if we leave anything to tests it might be perhaps linting, since it's so cheap (and just to defend in case we push stuff to the branch directly, for fixes, docs or anything).

tests/software/test_run_with_software.py
btraven00: thanks for the heads up, fixed!

.github/workflows/pipeline.yml
btraven00: thanks, removed

.github/workflows/pipeline.yml
btraven00: removed

.github/workflows/pipeline.yml
btraven00: so, one thing I'm realizing is that the more steps we have, the longer the overall pipeline will take
maybe does not do harm to have integration starting at the same time than shorts (with shorts being able to short-circuit and cancel the whole pipeline)

tests/workflow/test_run_workflow.py
btraven00: It seems to me that the intention is to verify that it's possible to define a benchmark yaml that defines only a single environment that is used by all steps. @imallona should be able to confirm.

this relates to some (missing) enforcement of not allowing heterogeneous backends; I think this was specified but not enforced.

In any case, in regard to testing, I think there's no need to actually test for real execution. Checking for some internal representation of the benchmark at a certain stage (i.e., pre-execution) should be enough. 

For now, we can either disable and defer this test (and related improvements), or keep it and defer.

.github/workflows/code-style.yml
btraven00: otoh, we can maintain this PR with a bunch of changes (docker image for software, other optimizations) and then remove the individual checks at the end.

I think the only point I'd like to discuss is I think we should run the same battery of tests after each merge (to main, dev).

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,264 @@
...
+    name: Code Style Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
imallona, line 22: worth keeping aligned with line 41? (via matrix/constant)?

imallona (approved):
Really awesome, thanks!

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,264 @@
...
+          pytest-xml-coverage-path: ./reports/coverage.xml
+
+  # Step 3: Run software tests only if regular tests pass
+  software-tests-linux:
+    name: Software Tests (Linux)
+    needs: tests
+    permissions: write-all
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
```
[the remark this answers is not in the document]
btraven00: In theory yes.
But I'm a bit hesitant to add one more platform to the tests until we can streamline the pipeline a bit more.

Do you think it's equally important to run short / integration / e2e on the matrix, or we can do with just the short tests?

tests/workflow/test_run_workflow.py
btraven00: removed, marking as resolved

DanInci (approved):

PR #98, opened 2025-05-02
chore: fix optional deps

btraven00:
## Description

Fixing two small bugs, one dealing with the incomplete s3 optional deps commit, and the other capturing the output of the conda/ etc checks

## Checklist:

## Remarks for the reviewer:

Some of the parameter typing that you see is trying to make type checking happy for the module (i.e., you cannot assign None to a string, you need Optional[str], and that messes with the positional vs keyword parameters. The other change you see here is related to overriding a string "benchmark" with a Benchmark object. reusing the same variable name confuses the checker and it complains; I tried to fix some of this while working on the module.

there's also a refactor of functions from the very generic "utils" into more appropriate module names.

### DIFF
--- a/omnibenchmark/io/hash.py
+++ b/omnibenchmark/io/hash.py
@@ -1,0 +1,10 @@
+import hashlib
+
+
+# https://stackoverflow.com/a/3431838
+def md5(fname: str):
+    hash_md5 = hashlib.md5()
+    with open(fname, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            hash_md5.update(chunk)
+    return hash_md5.hexdigest()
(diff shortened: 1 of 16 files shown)


omnibenchmark/io/hash.py
```diff
@@ -0,0 +1,10 @@
+import hashlib
+
+
+# https://stackoverflow.com/a/3431838
+def md5(fname: str):
+    hash_md5 = hashlib.md5()
+    with open(fname, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            hash_md5.update(chunk)
+    return hash_md5.hexdigest()
```
DanInci, line 10: Better name this function as checksum?

DanInci (approved):

omnibenchmark/io/hash.py
btraven00: agree, changed and force-pushed

PR #100, opened 2025-05-02
tests: make a conda binary available in the environment

btraven00:
According to the snakemake docs, a `conda` binary is needed in the environment:
https://snakemake.readthedocs.io/en/stable/getting_started/installation.html

snakemake 8.26 deprecated the use of conda-frontend (we were using mamba).

We need to update the docs and remove the reference to mac-test-environment.yaml


omnibenchmark/software/conda_backend.py
```diff
@@ -1,16 +1,19 @@
...
 
 
+@pytest.mark.conda
 def pin_conda_envs(benchmark_yaml: Path) -> None:
     """
     Pins a conda env file using snakedeploy
     """
     snakeconda.pin_conda_envs(
-        conda_env_paths=[benchmark_yaml], conda_frontend="mamba", warn_on_error=True
+        conda_env_paths=[benchmark_yaml], conda_frontend="conda", warn_on_error=True
```
imallona, line 18: how is this tested? I don't think snakedeploy will work with 'conda' as conda_frontend

PR #101, opened 2025-05-02
Automate docs via mkdocs

imallona:
This action automates the documentation generation to avoid the renku gitlab step.

- [ ] Update `settings -> Pages` to `gh-pages -> root` so the gh-pages rendered HTMLs are served.
- [ ] check the CNAME status with Mark / gandi


imallona:
This is an internal merge so I can test on omnibenchmark:omnibenchmark.

PR #103, opened 2025-05-02
docs: minor changes to readme after documentation site updates

btraven00:
## Description

- Minor updates to README
- Rename inner folder in docs to reduce confusion


btraven00:
hmm I think the publish mdocs task should not run for a PR, doesn't it?
or maybe it makes sense to run - to catch possible errors - but not push it?

imallona:
> hmm I think the publish mdocs task should not run for a PR, doesn't it? or maybe it makes sense to run - to catch possible errors - but not push it?

should we render when tagging instead?

imallona:
it does break indeed https://github.com/omnibenchmark/omnibenchmark/actions/runs/14833401999/job/41639377520#step:4:525

imallona (changes requested):
the `docs/docs` -> `docs/src` needs revision. Either rollback of modifying the `mkdocs` call.

it does break indeed https://github.com/omnibenchmark/omnibenchmark/actions/runs/14833401999/job/41639377520#step:4:525

btraven00:
> the `docs/docs` -> `docs/src` needs revision. Either rollback of modifying the `mkdocs` call.
> 
> it does break indeed https://github.com/omnibenchmark/omnibenchmark/actions/runs/14833401999/job/41639377520#step:4:525

thanks for catching that. fixed by specifying `docs_dir`

btraven00:
> > hmm I think the publish mdocs task should not run for a PR, doesn't it? or maybe it makes sense to run - to catch possible errors - but not push it?
> 
> should we render when tagging instead?

what would make sense to me is:

- run mkdocs build (normal) on every PR, as just another task. with this we catch that documentation compiles, and we can even preview with some extra work.
- run `mkdocs gh-deploy` only from blessed branches ( e.g. {dev, main})

if we agree on something along these lines I can incorporate in the ongoing pipeline-split work (for instance, we want to condition pushing docs to passing tests).

imallona:
have you considered switching to `mike` as well? https://github.com/jimporter/mike

btraven00:
mike looks nice, deferring to a new issue.

DanInci (approved):

PR #115, opened 2025-05-06
docs: add a readthedocs config

btraven00:
Configure readthedocs to support multiple branches and tags in the public-facing documentation.


btraven00:
My intent with this PR is try and see empirically if RTD can build the current documentation (for multiple branches) and then we decide if we go ahead with RTD or mike or other documentation platform.

Issue #109, opened 2025-05-05
pydantic import error on conda + python3.13


btraven00:
while attempting to run conda tests on the in progress test container, this error surfaces:

```bash
Activating conda environment: .snakemake/conda/12b2c6d7afc5f32deb514b9b8f86593a_
Traceback (most recent call last):
  File "/tmp/repo/tests/software/.snakemake/scripts/tmpfrsc11pm.run_module.py", line 12, in <module>
    from omnibenchmark.benchmark import constants
  File "/tmp/repo/omnibenchmark/benchmark/__init__.py", line 1, in <module>
    from .benchmark import Benchmark
  File "/tmp/repo/omnibenchmark/benchmark/benchmark.py", line 6, in <module>
    from pydantic import ValidationError
  File "/opt/conda/lib/python3.13/site-packages/pydantic/__init__.py", line 418, in __getattr__
    module = import_module(module_name, package=package)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/repo/tests/software/.snakemake/conda/12b2c6d7afc5f32deb514b9b8f86593a_/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.13/site-packages/pydantic_core/__init__.py", line 6, in <module>
    from ._pydantic_core import (
ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
```

btraven00:
In this particular case, the error is coming from a mismatch between the python interpreter in the conda environment (py3.13) and the interpreter installed in the conda environment managed by snakemake (3.12).

adding a explicit pydantic import in the test-environment.yml, and pinning python to `3.12` seems to be sufficient for, at least, the conda tests to pass.

btraven00:
this is solved by pinning pydantic like in https://gist.github.com/btraven00/297a4990c0a4bf7fdf591537804703dd

btraven00:
we pinned in dev, closing for now.

PR #119, opened 2025-05-08
chore: pin conda python deps

btraven00:
We had already pinned python3.12 in master (mostly due to the pydantic issue #109), so porting that change back to dev.

Additionally, I'm providing a convenience script to setup a conda environment with micromamba.


DanInci (approved):

scripts/create-micromamba-env.sh
imallona, line 1: I think this is a bad idea as we discussed, we don't know what's on the user's $PATH. I think it would be better to recommend installation via miniconda or miniforge.

scripts/create-micromamba-env.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# This script creates a micromamba environment called "omnibenchmark"
+# and installs the pinned system dependencies and an editable version of the
+# omnibenchmark python package.
+MAMBA=micromamba
```
imallona, line 5: micromamba is not friendly to neither mamba nor conda commands, I think this $MAMBA variable is confusing. We should perhaps assume conda is installed and just conda create the env; or suggest using miniconda or something more normal with a conda bundled in.

imallona: still, `test-environment.yml` is necessary for ci/cd tests for fast env / conda management, but then has the caveat of having to install `conda` (or `mamba`) inside.

for an user-facing recipe I'd go for miniconda or miniforge.

imallona (changes requested):
See comments above

btraven00:
I have reverted the instructions in the tutorial, and remove any reference to the script. For dev having a script is useful for me tho, since it saves some time. If we find it confusing to have it there I'm fine moving it somewhere else outside of the tree.

Since I was there, I removed a couple of things for the tutorial, like the mention to poetry, since it's not really needed anymore (in fact, python does install poetry as a build dependency if installing the folder, but there's no need to make it explicit, pip handles it underneath).

Re. changing the docs to use minimamba: maybe, once we're reasonably certain that it works. In any case, I'd just add a section as "recommended" way, and perhaps note that others work. We can do that in a new PR, but I'd make sure we switch the main tests to also use miniconda at the same time.

If any of  this is problematic, I'd rather split the PR into just the test-environment change and leave docs for a separate discussion.

scripts/create-micromamba-env.sh
btraven00: on the other hand, let's be aware that the script was isomorphic to the documentation :smiling_imp:

Issue #126, opened 2025-05-14
--continue-on-error flag is ignored


btraven00:
Running a module that raises an exception does not let the metric collector execute.

In fact, nothing else executes if the first task is a failing module.

To reproduce, I used the following dummy module:  https://github.com/btraven00/dummymodule

Which will raise a RuntimeException if passed any `--fail xxx` argument to the invocation.

Using dev @a5802516e

```bash
ob run benchmark --continue-on-error --local --task-timeout 4h --cores 10 -b Clustering_conda_smoketest.yml
S3 storage not available. You might want to install the 'minio' and 'boto3' packages.
Benchmark YAML file integrity check passed.
Are you sure you want to run the full benchmark even if some jobs fail? [y/N]: y
Running benchmark...
>>> run_workflow: timeout 14400
>>> args to api
Assuming unrestricted shared filesystem usage.
host: mlshainan
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 10
Rules claiming more threads will be scaled down.
Singularity containers: ignored
Job stats:
job                                                                                            count
-------------------------------------------------------------------------------------------  -------
all                                                                                                1
clustering_dummymodule_.71f1c7fc61e924fed27afc7d4369397d6a9a865e1dea72179974bf9894752972           1
metric_collector_plotting                                                                          1
metrics_partition_metrics_.1e99f1ac82179362b6fdc9caa9211fb8da6ba463b5e6078ef371eac70f88b788        1
metrics_partition_metrics_.dab8ec0285c698339308eb016fd5b021aa268af903e83b356c18f6ed076e4d95        1
total                                                                                              5

Select jobs to execute...
Execute 1 jobs...

[Wed May 14 13:54:59 2025]
localrule clustering_dummymodule_.71f1c7fc61e924fed27afc7d4369397d6a9a865e1dea72179974bf9894752972:
    input: out/data/clustbench/.46bb23e08cfbf0ba58db7fd92e04651ba558adc9a01ad0e7bacebda3866d7821/clustbench.data.gz, out/data/clustbench/.46bb23e08cfbf0ba58db7fd92e04651ba558adc9a01ad0e7bacebda3866d7821/clustbench.labels0.gz
[...]

PR #128, opened 2025-05-15
fix --continue-on-error flag (Closes: #126)

btraven00:
## Ticket

#126

## Description

This PR is mostly a fix for #126 

The result of my analysis is that the `--keep-going` flag was not working as intended.

The ultimate goal is to allow the fan-in part of the analysis continue even in the case of failures.

It also utilizes the testing infrastructure I had developed some time ago in #90 for embedding git buncles, so I'm closing that PR since it's now contained in this one.

However, 

## Checklist:

## Remarks for the reviewer:

I've split the commits in separate units, I think it makes sense to leave them rather than squash.


btraven00:
Tests are failing because timeut was reverted. Disabling them in #130

omnibenchmark/io/code.py
```diff
@@ -48,6 +51,7 @@ def get_git_archive_from_github(
 
 
 # https://docs.gitlab.com/ee/api/repositories.html#get-file-archive
+# TODO: deprecate, just use clone_git_repo
 def get_git_archive_from_gitlab(
     output_dir: Path, repository_url: str, commit_hash: str
 ) -> Path:
```
DanInci, line 57: Can we just delete this code if its deprecated?

omnibenchmark/utils.py
```diff
@@ -64,7 +64,10 @@ def merge_dict_list(list_of_dicts):
...
 def format_name(path, prefix):
     pattern = rf"{prefix}/.+?/([^/]+)/.+?$"
-    dataset = re.match(pattern, path)[1]
+    matched = re.match(pattern, path)
+    if matched is None:
+        return ""
+    dataset = matched[1]
     new_path = path.format(dataset=dataset)
 
     return new_path
```
DanInci, line 73: Can we add types here and document what kind of name is formatted?

DanInci (approved):

omnibenchmark/io/code.py
btraven00: as we discussed, the should be read as "to be deprecated". I'm not clear if the current cloning strategy is intended as optimization or what.

PR #132, opened 2025-05-20
feat: enable --task-timeout

btraven00:
we had temporarily disabled this to debug properly how it affects the execution of a complex benchmark.

we need to be sure not to have a regression of the parallelism being disabled by us adding custom resources that affect the DAG calculation.

## Ticket

## Description

## Checklist:

## Remarks for the reviewer:


omnibenchmark/workflow/snakemake/rules/rule_all.smk
```diff
@@ -14,19 +14,23 @@ import logging
...
     else:
         rule all:
             input: paths
             output:
                 f"{benchmark.out_dir}/performances.tsv"
+            params:
+                local_task_timeout = local_task_timeout
             run:
                 result = glob_wildcards(str(benchmark.out_dir) + "/{path}/{dataset}_performance.txt")
                 performances = expand(str(benchmark.out_dir) + "/{path}/{dataset}_performance.txt", path=result.path, dataset=result.dataset)
```
DanInci, line 36: Why is a local task timeout needed here?
In Snakemake, rule_all is a special rule that defines the final output targets of the workflow. It doesn’t do any actual computation itself; rather, it serves as the goal for the DAG resolution. Sure, there's the actual else case where there's the computation of concatenating performance results, but that's negligable.

omnibenchmark/workflow/snakemake/scripts/execution.py
```diff
@@ -67,6 +67,7 @@ def execution(
     stderr_file = output_dir / "stderr.log"
 
     logging.info(f"Seting timeout to {timeout} seconds")
+    print(">>> timeout", timeout)
```
DanInci, line 70: Remove `print`

omnibenchmark/workflow/snakemake/rules/rule_all.smk
DanInci: This is anyways not picked by the `execution.py` script.

omnibenchmark/workflow/snakemake/snakemake.py
```diff
@@ -39,7 +39,7 @@ def run_workflow(
         module_path: str = os.environ.get("MODULEPATH", ""),
         debug: bool = False,
```
DanInci, line 40: You probably want to edit the documentation and method definition in the interface `WorkflowEngine`

omnibenchmark/workflow/snakemake/rules/rule_all.smk
btraven00: the logic was to apply the timeout also to the aggregation task, but I'm fine removing it

omnibenchmark/workflow/snakemake/snakemake.py
btraven00: done, thanks

Issue #106, opened 2025-05-05
allow to override the --out folder


btraven00:
we should be able to override `out` folder

PR #133, opened 2025-05-20
Add --out-dir flag

DanInci:
## Ticket

https://github.com/omnibenchmark/omnibenchmark/issues/106

## Description

- [x] Add --out-dir flag

## Checklist:

## Remarks for the reviewer:

The `--out-dir` option is now restricted to use only with the `--local` flag. I added this constraint because the current storage/io logic is complex enough and expects workflow results to be always stored in `out`. Given this, archival, versioning, and listing storage files all expect the results to be in `out`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 This document records all notable changes to `omnibenchmark`.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
-## [0.2.2] UNRELEASED
+## [0.2.2](main) UNRELEASED
 - cli: added `--out-dir` to `ob run benchmark` to configure a different output directory for the workflow
 
 ## [0.2.1](https://github.com/omnibenchmark/omnibenchmark/releases/tag/v0.2.1) (May 20th 2025)


retogerber:
Just checked the archiving where the --out-dir flag is still missing so currently it wants to archive results under output directory "out". I think here the storage_options object needs to be passed https://github.com/omnibenchmark/omnibenchmark/blob/e6a3d3f018e4309e55ddfe6480a363f54e2a5bf9/omnibenchmark/io/archive.py#L158

```
ob run benchmark -b tests/data/Benchmark_001.yaml -l --out-dir out2
ob storage archive -b tests/data/Benchmark_001.yaml -l -r -n
```
```
Not all expected files are available locally
Missing files:
out/data/D1/default/process/P1/.47fae0c4eff9d60d70f80ba26f299ece136010217b7393c672091bfbd5cecb1c/methods/M1/default/D1.model.out.gz
[...]
```

DanInci:
@retogerber would it make sense to add to all storage commands an --out-dir flag?

DanInci:
@retogerber That's what I was thinking by constraining the `--out-dir` to be allowed only with the `--local` flag. Then for the archiving, it would still be needed to specify which results folder you want to archive (and by default is `out`). @imallona @btraven00 any takes on this?

btraven00:
I don't think the main "out" folder needs to be reflected in the remote storage "folder" hierarchy abstration; semantically, it does not make sense, I would archive just the contents of the out folder. I think that solves the problem, am I missing anything?

imallona (approved):
I think the outdir specification should work for both remote and local runs to make the CLI easier to predict/use. This is just a preference, the PR looks awesome.

Issue #140, opened 2025-06-10
define module validation spec


btraven00:

btraven00:
As reference, [here is a link to the previous module validation spec](https://github.com/omnibenchmark/internal_docs/blob/master/architecture/method_contributor_design.md) I was working with

Issue #104, opened 2025-05-03
consider using CFF as module metadata


btraven00:
it seems github and zenodo have some sort of support for CFF.

This could simplify a bit the collecting of citations for a benchmark, I think (recursive parsing of all modules).
https://citation-file-format.github.io/

How do we make all modules have a valid citation is perhaps another story :)

imallona:
Looks interesting. Very needed to collect and report these during benchmark runs/report generation, as well as to validate the licenses.

imallona:
Is this complete as in "it has been discussed and design doc proposal written" or as in "it is implemented and we get a BibTeX record or some log/reporting to attribute a benchmark run"? 

By attribute I mean whether `ob` collects the CFFs when running a set of modules specified within a benchmark manifest and generates a report listing authors. @btraven00

PR #145, opened 2025-06-13
docs(design): add module contributor spec for discussion

btraven00:
## Ticket

- https://github.com/omnibenchmark/omnibenchmark/issues/140
- https://github.com/omnibenchmark/omnibenchmark/issues/104

## Description

Please read [the design document in the PR](https://github.com/omnibenchmark/omnibenchmark/pull/145/files?short_path=07689e8#diff-07689e8bd64b764eb602edf1e4a0016c7874a681b707e36c093e045915154763) for the full details. Feel free to react to this issue with emojis for loose agreement, add a proper issue review, and/or use the comments. 

## TLDR: Split Module Metadata into Two Files

We propose using a hybrid metadata layout for contributed modules:

- `CITATION.cff`— for all attribution/citation metadata (see #104). The plan would be to expose this functionality in a `omnibenchmark generate-citations` command in the future (it would simply recursively collect all the contributed citations. Benchmark validators can also hook into checking of metadata (e.g. valid licenses, valid authorship).
- rename `config.cfg` to `omnibenchmark.yaml` — for _execution_ _semantics_: entrypoints, environments, etc. and any further extensibility on the side of module execution environments.

## Why this split?

- `CITATION.cff` is widely supported by GitHub, Zenodo, and citation tooling. No need to reinvent attribution.
- Keeping a module-level `omnibenchmark.yaml` focused on execution (entrypoints, conda, EasyBuild, etc.) avoids overloading the CFF schema

This separation mirrors best practices (e.g. pyproject.toml + CITATION.cff for python packages), even if some level of duplication exists between the packaging metadata and the cff format (but they serve different purposes).

### DIFF
--- a/docs/design/001-module-metadata.md
+++ b/docs/design/001-module-metadata.md
@@ -1,0 +1,156 @@
+# 001: Module Metadata
+
+[![Status: Approved](https://img.shields.io/badge/Status-Approved-green.svg)](https://github.com/omnibenchmark/docs/design)
+[![Version: 0.1](https://img.shields.io/badge/Version-0.1-blue.svg)](https://github.com/omnibenchmark/docs/design)
+
+**Authors**: ben, ...
+**Date**: 2025-06-18
+**Status**: Approved
+**Version**: 0.1.0
+**Supersedes**: N/A
+**Reviewed-by**: dincicau, csoneson, imallona
+**Related Issues**: #145
+
+## Changes
+
+| Version | Date | Description | Author |
+|---------|------|-------------|--------|
+| 0.1 | 2025-06-13 | Initial draft | ben |
+
+
+## 1. Problem statement
+
+Previous [spec for module metadata](https://github.com/omnibenchmark/internal_docs/blob/master/architecture/method_contributor_design.md) suggested to expose attribution fields (for citation) and execution details (entrypoints, environment specs) in a single YAML file.
+
+We also rely on a `config.cfg` file to be [present at the root of a valid omnibenchmark module](https://github.com/imallona/clustbench_fcps/blob/main/config.cfg). As of today, this file only contains information about the module entrypoint:
+
+```
+[DEFAULT]
+SCRIPT=run_fcps.R
+```
+
+It has also been noted in the past that the `1:1` relation between module and entrypoint leads to a perhaps unnecessary proliferation of repos. As a secondary goal, being able to expose more than one entrypoint could allow to reuse a repo for different benchmark stage (it's just a namespace.)
+
+
+### Design Goals
+
+- Leverage existing tooling (GitHub, Zenodo) for citation.
+- Lower the barrier to entry for new contributors.
+- Keep execution metadata concise, clearly separated.
+- Support extensions (new entrypoints, env types) without bloating either file.
+
+## 2. File Roles
+
+Current spec expose both citation (attribution) metadata and execution details (entrypoints, environment specs) in a single YAML. We propose moving to a hybrid spec:
+
+1. `omnibenchmark.yaml` for module execution semantics (entrypoints, named stages, environment specs: conda, easyconfig).
+
+```YAML
+entrypoints:
+  default: shuffler.run
+environments:
+  conda: this-environment.yaml
+  easybuild: https://…/AMGX-2.4.0.eb
+```
+
+2. Adopt a `CITATION.cff` alongside it for all citation/attribution fields (title, authors, abstract, references, license).
+
+```YAML
+cff-version: 1.2.0
+message: "If you use this module in your publication, please cite as below."
+title: "Shuffler"
+abstract: |
+  Shuffles (per row) a rectangular gz-compressed, tab-separated text file.
+  Known to break on embedded quotes. Reruns with the same seed are identical.
+authors:
+  - family-names: Turtle
+    given-names: John
+date-released: 2025-06-13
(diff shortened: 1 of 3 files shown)


csoneson (approved):
I have read the design document and I agree with the proposed hybrid approach. The possibility to specify multiple entrypoints and environment definitions will likely also be useful - should these be used as fallbacks if they are not explicitly specified in the benchmark definition? 

Regarding the tool to create cff files, there is https://citation-file-format.github.io/cff-initializer-javascript/#/ that can perhaps be used.

DanInci:
Looks really good to me, only a couple of considerations:
- we don't want to introduce unnecessary hassle to contribute a module, hence we should define the exact minimum requirements for both `omnibenchmark.yaml` and `CITATION.cff`, an example of the minimum for us to agree on would be great (i.e. entrypoints, author, license, ...) ?
- `environments` keyword might be ambigous, given that they are used to check, rather then enforce a particular env, i'd rename it to something like `environment_constraints` 
- Also, do we want to introduce versioning for the `omnibenchmark.yaml` ? In that case i'd add a `version` keyword there

DanInci (approved):

imallona:
Looks good! Re: the entrypoints and the 

> It has also been noted in the past that the `1:1` relation between module and entrypoint leads to a perhaps unnecessary proliferation of repos.

I don't fully get it, how would `ob` decide which entrypoint to run from the `1 module : n entrypoints`? Related, is this feature (having one module offering different entrypoints) user friendly?

docs/design/001-module-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+- `omnibenchmark.yaml` can adopt new environments types as they are needed (e.g. Docker, Singularity).
+- `omnibenchmark.yaml` can define new entrypoints as they are needed.
+
+
+## 6. Recommendations and next steps
+
+- Adopt the hybrid approach.
+- Make a bare minimum of fields mandatory (entrypoint, URL, author, LICENSE)
+- Update docs to guide module contributors, including best practices for generating citation metadata (e.g. using online tools).
+- Explicitely address differences between metadata about code, algorithms, datasets and publications.
```
imallona, line 147: I don't understand this, are there differences? In principle all reused code/data share some common patterns, e.g. license/authors/doi (if any)

docs/design/001-module-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+## 2. File Roles
+
+Current spec expose both citation (attribution) metadata and execution details (entrypoints, environment specs) in a single YAML. We propose moving to a hybrid spec:
+
+1. `omnibenchmark.yaml` for module execution semantics (entrypoints, named stages, environment specs: conda, easyconfig).
+
+```YAML
+entrypoints:
+  default: shuffler.run
+  postprocess: shuffler.verify
```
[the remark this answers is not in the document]
btraven00: agree. I will remove any other entrypoints other than "default". I still think this is worthy to discuss, but deferring it. fwiw, it was just an example of how/why why might want to extend the yaml semantics.

btraven00:
>     * we don't want to introduce unnecessary hassle to contribute a module, hence we should define the exact minimum requirements for both `omnibenchmark.yaml` and `CITATION.cff`, an example of the minimum for us to agree on would be great (i.e. entrypoints, author, license, ...) ?

I agree - at a minimum, I think author attribution and LICENSE - plus the entrypoints file. 

>     * `environments` keyword might be ambigous, given that they are used to check, rather then enforce a particular env, i'd rename it to something like `environment_constraints`

I am not 100% clear on what's the purpose of these entrypoints. In the document I mentioned these are informative and not notrmative, which is my current understanding. I would leave it as-is tbh, until we see if we make it actionable or not.

>     * Also, do we want to introduce versioning for the `omnibenchmark.yaml` ? In that case i'd add a `version` keyword there

I think it's unnecessary complexity; for the time being I would just rely on git for versioning. I see your point, but I think we need to improve schema evolution in a better way generally.

docs/design/001-module-metadata.md
btraven00: my thinking was that named entities might be different for these cases. but yes. in essence for reusing a pipeline I think one wants to programmatically annotate what's code, what's data, and what are different implementations of the same method, in different languages libraries etc

btraven00:
Updated the proposal:
- remove reference to more than one entrypoint
- qualify suggestion about distinctions between named entities regarding their role

imallona (approved):
I agree with Daniel adding a version field is going to be helpful and would recommend limiting entrypoints to only one. These are only preferences, the PR looks good to me.

btraven00:
> a version field is going to be helpful 

last Tuesday, while debugging broken osx tests, I was chatting with daniel about this point. It'd be good to clarify, a version of what. Right now the spec version is not coupled to omnibenchmark version. That will change when we remove omni-schema. In that moment, you want a version to (a) track changes to the yaml itself (for that we have git, branches and tags); and (b) track changes to the spec. This spec version will match the omnibenchmark version. If it's true that we're following semver, then only major changes will break API; all the rest will forcefully be backwards-compatible (when we reach some stability, of course). So yes, we'll want to track major spec version compatibility, but this will only matter when dealing with incompatible versions of omnibenchmark; preparing for that, we can put in place a mechanisms for schema migration (south style) and attempt automatic migration.

Issue #137, opened 2025-05-27
Support for HPC / SLURM is missing


imallona:
Running `ob` with a queuing system is expected. 

Worth designing the feature with at least these aspects in mind:
- using job groups/array jobs (given the thousands of jobs each trying to reuse some envs)
  - to ease software management (e.g. with `conda` for jobs sent to multiple nodes)
  - to ease input/output files chaining
- profiling resources
  - and even allocation, e.g. keeping the code open to future modifications to send jobs to different nodes (`highmem`, `fast`, etc) 
- sending env vars and particularly secrets/credentials (e.g. for writing to S3)

Worth discussing how to this to the YAML or to the `ob` call as extra snmk invocation args (#129) or something else.

btraven00:
Assigning to @DanInci after today's team meeting discussion on priorities

PR #147, opened 2025-06-17
feat: SLURM support and custom Snakemake executors

DanInci:
## Ticket

https://github.com/omnibenchmark/omnibenchmark/issues/137

## Checklist:

## Remarks for the reviewer:

Tested on the science IT cluster. Used the following script, which successfully generated expected results
```bash
#!/usr/bin/bash -l

#SBATCH --time=0-00:10:00    # 10 minutes walltime to cover longest Snakemake job
#SBATCH --mem=4G             # 4 GB RAM
#SBATCH --ntasks=1           # number of tasks (usually 1 for most jobs)
#SBATCH --cpus-per-task=4    # number of CPUs per task

module load mamba rstudio/4.3.1
source activate omnibenchmark

ob run benchmark -b tests/data/Clustering.yaml --local \
  --executor slurm \
  --jobs 6 \
  --cores 24 \
  --default-resources mem_mb=4000 runtime=600 \
  --verbose --printshellcmds
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - feature: SLURM support
 - feature: Support passing of extra arguments from CLI -> Snakemake for run commands
 
+## [0.2.2](https://github.com/omnibenchmark/omnibenchmark/releases/tag/v0.2.2) (Jul 16th 2025)
+
+- feat(cli): re-add temporarily disabled --task-timeout (#132)
+- feat(cli): added `--out-dir` to `ob run benchmark` to configure a different output directory for the workflow
+- feat: add general config file (#148)
+- docs: clarify documentation on install methods (#144)
+- docs: add design docs template
+
 ## [0.2.1](https://github.com/omnibenchmark/omnibenchmark/releases/tag/v0.2.1) (May 20th 2025)
 
 - bugfix for `--continue-on-error` flag.
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,75 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: omnibenchmark
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: 'Izaskun '
+    family-names: Mallona
+    orcid: 'https://orcid.org/0000-0002-2853-7526'
+  - given-names: Almut
+    family-names: Lütge
+    orcid: 'https://orcid.org/0000-0003-1104-727X'
+  - given-names: Ben
+    family-names: Carrillo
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -5,7 +5,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [0.2.2](main)
 
-- ...
+- feature: SLURM support
```
btraven00, line 8: since this is a (mayor) new feature, I would bump to 0.3.0

tests/cli/test_run.py
```diff
@@ -90,3 +90,41 @@ def test_run_benchmark_with_yes(
...
+
+    # Ensure workflow.run_workflow is called
+    mock_workflow_run_workflow.assert_called_once()
+    args, kwargs = mock_workflow_run_workflow.call_args
+    assert kwargs["cores"] == 1
+    assert kwargs["executor"] == "slurm"
+    assert kwargs["jobs"] == "2"
+    assert kwargs["continue_on_error"] is True
+
+    assert result.exit_code == 0
```
btraven00, line 130: great. don't we need to also configure auth etc in the profile? if so we might need to worry about testing also precedence/overrides...

omnibenchmark/cli/utils/args.py
```diff
@@ -0,0 +1,26 @@
+def parse_extra_args(args):
```
btraven00, line 1: this seems a bit brittle to me. is this the recommended way? afaik click supports passing extra args, with n args explicitly, in the form -e foo bar. Isn't that enough? I'm worried this might break in unexpected ways (for instance mesing with ordering of the flags)

but not a strong objection, if it's well tested and solves the problem, why not. I would avoid the "utils" catch-all submodule tho :grimacing:

omnibenchmark/workflow/workflow.py
```diff
@@ -21,11 +21,12 @@ def run_workflow(
...
         keep_module_logs: bool = False,
         backend: SoftwareBackendEnum = SoftwareBackendEnum.host,
-        module_path: str = os.environ.get("MODULEPATH", None),
+        module_path: str = os.environ.get("MODULEPATH", ""),
+        debug: bool = False,
         work_dir: Path = Path(os.getcwd()),
         out_dir: str = "out",
         resources: Optional[dict] = None,
-        **kwargs,
+        **snakemake_kwargs,
```
btraven00, line 29: minor (and this was bound to happen), but this is the interface declaration, not the actual workflow implementation. Yes, I know we're into tight coupling, but all the same it was a nice idea.

btraven00 (approved):
I left minor comments, mostly for discussion. But if it works, it's a sufficiently minimal example!

btraven00:
> @btraven00 Agree, from the UX perspective, the `--local` flag is confusing. Alternatives I'm thinking of:
> 
>     * `--local-storage`

I think I would invert the logic: local by default, and then pass --archival s3 to enable

imallona (commented):
Are tutorial/docs updated? It would be good to know where/how to set up the auth.

Issue #31, opened 2024-10-25
Implement system-wide configuration and cache for ob command


DanInci:
Currently, there's no place to store system-wide configuration for omnibenchmark.
In addition, we could use some system-wide cache for repos, artifacts and other future needs of the project. (At the moment, repos are cached in the project specific .snakemake project)

PR #148, opened 2025-06-18
feat: add general config file (Closes: #31)

btraven00:
## Ticket

#31 

## Description

- Add general configuration accessor
- Document configuration file location and format
- Replace hardcoded paths with config-based lookups

## Checklist:

## Remarks for the reviewer:

Here we can configure snakemake specific settings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,6 +133,18 @@ This ensures two things:
 
 We might change this in the future to a more generic approach. But for now this should be sufficient to allow debuggability of the test outputs, and good test isolation.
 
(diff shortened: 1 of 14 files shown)


imallona (approved):
I haven't tested this nor checked the code thoroughly. Still it looks great and needed. A couple of thoughts (they're not questions to be replied to): What about S3-, singularity registries- and other secrets? And slurm settings? Similarly, and snakemake profiles? I also wonder how to do the section structuring in an extendable way: e.g. modulepaths, for machines where software is not managed by easybuild but envmodules exist and are used.

Issue #150, opened 2025-06-18
add a more robust profiler to the benchmark tasks


btraven00:
The current profiling in omnibenchmark is brittle: shows no mem, io metrics for methods.
In part this might be caused by very short runtimes, but in part it might be something else.

# Definition of done

omnibenchmark is set to use and output alternative monitoring metrics for the executed tasks. 

## Out of scope

Processing of the outputs

imallona:
For documentation/logging purposes: https://github.com/btraven00/denet/ has been released by btraven to address this

PR #151, opened 2025-06-23
feature: add extra profiler to the snakemake execution

btraven00:
## Ticket

#150 

## Description

Add an extra profiler to have a second source of information about process execution. This profiler only executes if the dependency is detected in the environment.

## Checklist:


omnibenchmark/workflow/snakemake/scripts/execution.py
```diff
@@ -70,15 +78,28 @@ def execution(
...
+                # execute and monitor
+                exec_fn = execute_with_monitoring
+                kwargs["stdout_file"] = stdout_file
+                kwargs["stderr_file"] = stderr_file
+                kwargs["output_file"] = (output_dir / "perf_raw.jsonl").as_posix()
+                exit_code, monitor = exec_fn(command, **kwargs)
+                # output the aggregated metrics in the folder too
+                with open((output_dir / "perf.json").as_posix(), "w") as perf_file:
+                    summary = json.loads(monitor.get_summary())
+                    json.dump(summary, perf_file)
```
DanInci, line 93: Can we write in the dump what profiler we are using?
So it's self documented that we are using `denet`.

PR #154, opened 2025-07-10
design: add module artifact validation proposal

btraven00:
## Ticket

## Description

## Checklist:

## Remarks for the reviewer:


DanInci (commented):
Thanks ! The proposal looks solid and will definitely work !

However, I have 2 concerns about it:
**1. Having a stage level validations implies unified output structure:** but that might not hold true in our setup. For example, we're still using the `exclude:` keyword, which introduces mutually exclusive paths. This means that two methods could produce the same file name with entirely different structures, and each would be intended for a different processor in the next stage. Validating them uniformly at the stage level could lead to incorrect assumptions or failures.

`2. Reinventing the wheel with the Mini-DSL:` I'm not a fan of building and creating new DSL for validation. But I do agree with your take that json schema and CSV schema can be quite complex to learn. Have you checked out [DBT schema and python models](https://docs.getdbt.com/docs/build/python-models)? Seems to be simple enough and the project is mature and used for data engineering pipelines.

PR #159, opened 2025-07-22
refactor: remove the software cli subcommand

btraven00:
it was agreed this functionality does not belong on the main ob cli command.

PR #160, opened 2025-07-22
fix(tests): disable failing apptainer test, deprecated

btraven00:
This test is failing in main, but we're going to disable it in the next release (#159), so disabling it for now so that we can tag 0.2.2.


DanInci (approved):

DanInci (changes requested):
Is this still needed given that we're removing the software CLI commands and the cache invalidation?

imallona (approved):

btraven00:
>  Is this still needed given that we're removing the software CLI commands and the cache invalidation?
no, 0.2.2 is already tagged. closing.

PR #169, opened 2025-08-08
chore: release 0.3.0

btraven00:
PR to main, containing current state of dev for 0.3.0 tag.


btraven00:
I would like to suggest switching to squash commits as default for PRs to dev for next dev cycle. The history is a bit messy.

DanInci (approved):

docs/src/howto.md
```diff
@@ -1,3 +1,229 @@
...
+
+Omnibenchmark is a pip-installable python package ([PyPI](https://pypi.org/project/omnibenchmark/), [source code](https://github.com/omnibenchmark/omnibenchmark)).
+
+### Supported Platforms
+
+| Backend      | Linux | MacOS | Windows |
+|--------------|-------|-------|---------|
+| Conda        | ✅    | ✅    | ❌       |
+| Apptainer    | ✅    | ❌    | ❌       |
+| Easybuild    | ✅    | ✅    | ❌       |
```
imallona, line 11: is Easybuild truly supported on MacOS? most easyconfigs won't work there

docs/src/tutorial.md
```diff
@@ -606,7 +132,7 @@ stages:
         ##   params: `empty` (no parameters added)
```
imallona, line 132: Lines 128-132 are not needed anymore

imallona (approved):

imallona:
If the changelog does not report the CLI arguments renaming (local-storage and perhaps others) I think it would be good to include it, also as a general rule

btraven00:
merged and tagged! :partying_face: :balloon:

Issue #171, opened 2025-08-22
add templaes for benchmark & module creation


btraven00:
We want a `ob create benchmark` and a `ob create module` command, to facilitate scaffolding of new contributions. 

Previous attempt at templating: https://github.com/omnibenchmark/omnibenchmark/pull/59 

https://copier.readthedocs.io/en/stable/ seems to be a very nice option in the python ecosystem - it will allow us to, for instance, update templates against a URL.

For now I will embed the templates within the repo (and the python package), but we can do online and even a template per benchmark - thus giving freedom to the benchmarker to have their own templating logic.

PR #172, opened 2025-08-22
feat: ob create benchmark & module (Closes: #171)

btraven00:
## Ticket

#171

## Description

Add basic capabilities to create new benchmark and modules, using  pre-defined templates.

Adding copier as the main engine for templating - uses jinja templates.

## Checklist:

## Remarks for the reviewer:

More interested in the dependencies and process than on feature completeness. We can refine later for 0.4.0


omnibenchmark/templates/module/README.md.jinja
```diff
@@ -0,0 +1,137 @@
...
+```
+
+## License
+
+This project is licensed under the {{ license }} License - see the `CITATION.cff` file for details.
+
+## Support
+
+- [OmniBenchmark Documentation](https://omnibenchmark.org/docs)
+- [Report Issues](https://github.com/CHANGE_ME/{{ module_name }}/issues)
```
imallona, line 133: hmm, `CHANGE_ME` not in jinja for obvious reasons. worth removing the line altogether?

btraven00: I was leaving it there so that we can fail the validation, to remind people that they need to add a link to their public remotes.

btraven00:
I would not bother fixing these tests if they're gonna be revamped in #170

btraven00:
actually the yaml reading will need adaptation after merging #170 (do we want to land that today too?)

imallona:
agree. It passes the 'clustbench' integration test. Let me install and check the commands themselves...

imallona (changes requested):
Please make sure `ob run benchmark` runs with the license without breaking the validation, see comment above.

Otherwise looks great

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 This document records all notable changes to `omnibenchmark`.
 This project uses [Semantic Versioning](https://semver.org/) and [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
 
+## [0.4.0](main) UNRELEASED
+
+- feat: add create benchmark & create module command (#171)
```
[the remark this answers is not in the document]
btraven00: (if we want to release 0.3.1 for whatever reason I think it should be cut from dev into main as cherry-picks, or via a detached state from the release/0.3.0 branch)

btraven00:
fixed the license stuff, and one bug on the created identifiers.

ob info now correctly parses the topology:

```
❯ ob create benchmark
Starting copier questionnaire for new benchmark using OmniBenchmark v0.3.0.post10+dirty...
🎤 What is the name of your benchmark?
   my-awesome-benchmark
🎤 What is your name?
   Your Name
🎤 What is your email?
   you@example.com
🎤 Choose a license for your benchmark
   MIT
🎤 Brief description of your benchmark
   A brief description of what this benchmark does
🎤 Choose the default software environment backend
   conda

Copying from template version None
    create  CITATION.cff
    create  validations.yaml
    create  envs
    create  envs/conda.yaml
    create  .gitignore
    create  README.md
    create  benchmark.yaml


Initialized Git repository
Created initial commit
Benchmark scaffolding created successfully with OmniBenchmark v0.3.0.post10+dirty
Next steps:
  1. Review and customize the benchmark configuration
  2. Add your benchmark modules
  3. Run 'ob validate' to check your configuration (coming soon!)
❯ ob info topology -b benchmark.yaml
---
title: my-awesome-benchmark
---
flowchart LR
	classDef param fill:#f96
	subgraph datasets
		dataset_loader
	end
	subgraph methods
		cool_method_1
		dataset_loader --> cool_method_1
	end
	subgraph metrics
		strict_metric_1
		cool_method_1 --> strict_metric_1
	end
	subgraph params_dataset_loader
		426fc04f04bf8fdb5831dc37bbb6dcf70f63a37e05a68c6ea5f63e85ae579376['--foo', 'bar']
	end
	params_dataset_loader:::param --o dataset_loader
	subgraph params_cool_method_1
		426fc04f04bf8fdb5831dc37bbb6dcf70f63a37e05a68c6ea5f63e85ae579376['--foo', 'bar']
	end
	params_cool_method_1:::param --o cool_method_1
	subgraph params_strict_metric_1
		426fc04f04bf8fdb5831dc37bbb6dcf70f63a37e05a68c6ea5f63e85ae579376['--foo', 'bar']
	end
	params_strict_metric_1:::param --o strict_metric_1
```

> I tried to overwrite the module starting once with an R entrypoint and next a bash, and it warned me about overwriting the README, omnibenchmark.yaml, citation etc; but it didn't delete the (deprecated) R entrypoint. I guess this is not unexpected, but feels odd.

Yea, I feel very unsafe to try to delete files. This should be a one-off operation.

>  I find a bit annoying you have to run it on an empty folder and then it tells you you need to cd there. You are there.

Fixed, doesn't hint to cd if target == cwd.

FWIW you can also instantiate passing a path: create benchmark /foo/bar

> I find the input and output arguments used for parsing useful for copypasting/extending, but I guess users (Mark) should be prepared and know these should match the benchmarking yaml's and not be input nor output. If ensuring some arguments are provided for parsing purposes, I think current ob requires a module_name and output_dir (perhaps named differently) that are mandatory arguments no matter the module's nature. Perhaps worth templating those?

I think we'd be better off by trying to design a good contract for what modules are expected to receive. I am not even sure out_dir is a generalizable abstraction - I feel we're designing the grammar too closely to snakemake at this point.

The whole argument parsing is very ad-hoc at this point, it might be that we do not need it at all if we pass arguments via the snakemke object.

Shall we try to discuss the mandatory arguments and the rules to generate the mapping in a different issue at this point?

btraven00:
> If ob create benchmark license was set to MIT by default, now it is set as GPLv3. Is this intended? And, in the future when running ob create module --bechmark yaml in the future, will it enforce the same license?

For clarity, I "fixed" that it's choosing the same default no (MIT) for benchmark and module.

Whether the module has to follow the same license tan the benchmark: I think the easiest thing to do is not to alow input, if given a reference benchmark definition. I am not sure it's legally problem, but itsure simplifies things to have all modules abiding by the same license (it's tricky because we're in the same situation as debian packaging etc: is it the license of the contribution, or the license of the software?).

At the metadata level, I would only care about consistency between the different possible fields (CFF, LICENSE file if written too etc). Seems more of a task for the validator.

imallona (approved):
Thanks for fixing the license so quickly! Lgtm

omnibenchmark/templates/benchmark/envs/{% if software_backend == 'conda' %}conda.yaml{% endif %}.jinja
```diff
@@ -0,0 +1,9 @@
+name: {{ benchmark_name | lower | replace(' ', '_') | replace('-', '_') }}_env
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python>=3.9
+  - pip
+  - pip:
+    - # Add your pip dependencies here
```
DanInci, line 9: Surprised this is an acceptable if statement in the file name. Looks so weird.

omnibenchmark/templates/module/{{ entrypoint }}.jinja
```diff
@@ -0,0 +1,107 @@
...
+import sys
+from pathlib import Path
+
+# Add src directory to Python path
+src_dir = Path(__file__).parent / "src"
+sys.path.insert(0, str(src_dir))
+
+from main import process_data
+
+def parse_args():
```
DanInci, line 14: Why do we keep both the `entrypoint.py` and the `src/run.py` as a template? 
I feel like there should be just one entrypoint to not add more confusion. The user / module contributor should be able to structure their own project, they will need just the wrapper for their method.

omnibenchmark/cli/create.py
```diff
@@ -0,0 +1,615 @@
...
+        if not git_dir.exists():
+            subprocess.run(
+                ["git", "init"], cwd=target_path, check=True, capture_output=True
+            )
+            logger.info("Initialized Git repository")
+
+        # Add all files and create initial commit
+        subprocess.run(
+            ["git", "add", "."], cwd=target_path, check=True, capture_output=True
+        )
```
DanInci, line 61: We still have the dependency on `"GitPython>=3.1.0"`. We should be consistent, and either use it, or remove it entirely from the codebase.

DanInci (commented):
Really great work ! Both commands are intuitive to use. 

I experienced just one issue with the software backends. Whenever choosing a `conda` environment for the benchmark / module, everything works fine, because a template conda.yml is created.

However, that's not the case with apptainer, or envmodule. The generated benchmark doesn't pass validation because its missing the backend file. I suggest at least touching the required file, if templating is not possible.

```Error: An unexpected error occurred: Software environment path for 'apptainer' does not exist: 'test/envs/example.sif'.```

omnibenchmark/cli/create.py
btraven00: done

PR #179, opened 2025-10-21
tests: add e2e tests

btraven00:
This PR provides end-to-end tests to gain assurance about overall outputs, given a different series of YAML descriptions.

Intended to give high-level quality checks for the #170 refactor, but we could also curate a reference list of YAMLs with different topologies, as suggested by @imallona during the review of #170


DanInci (approved):
LGTM

PR #185, opened 2025-11-06
docs: Copyedit and update user-facing documentation

imallona:
Attempt to copyedit, simplify and update before DK.

Rendered version: https://omnibenchmark--185.org.readthedocs.build/185/

Notes:
- commented out the configuration file `config.md` hence removing it from the navbar and the docs. Given that, apparently, it doesn't materialize as a file with current `ob`; also see feedback by Mark
- drafted an `examples.md` section
- added DK without any context to `news.md`
- removed names from the `team.md` and left only Mark

### DIFF
--- a/docs/src/news.md
+++ b/docs/src/news.md
@@ -2,9 +2,15 @@
 
 - Community benchmarking event at the University of Copenhagen. Onsite 17-18 November and asynchronous afterwards
 
+### September 2025
+
+- Omnibenchmark was present at _Basel Computational Biology Conference_ in Basel, Switzerland:
+  - [Omnibenchmark for reproducible and collaborative benchmarking (poster)](https://zenodo.org/records/17579265)
+
+
 ### March 2025
 
-- Omnibenchmark was part of the _Modern Benchmarking_ conference in Ascona, Swizerland:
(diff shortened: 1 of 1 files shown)


docs/src/tutorial.md
```diff
@@ -60,7 +58,7 @@ storage_api: S3
 ##   not sharing results from our benchmark yet.
 storage: https://s3_object_storage_url.ch
 
-## Benchmark YAML schema/specification version. Currently `0.01`.
+## Benchmark YAML schema/specification version.
 benchmark_yaml_spec: 0.01
```
btraven00: I think we should add strings in these examples for things like versions

btraven00 (approved):
I wonder how much this conflicts with the current state of docs in dev. In there for instance I added the automatic template etc.

I think it's good to merge this in main now, and then immediately after do a merge of docs/ into dev to deal with any conflicts, so that we prepare the merging back of dev into main after 0.4.0

We could also configure RTD to show latest for docs rendered from dev, but not urgent for now.

markrobinsonuzh (approved):

DanInci (approved):

Issue #189, opened 2025-11-06
unknown flags result in confusing usage message


btraven00:
Since we started passing extra arguments verbatim to snakemake, the use of certain flags results in somewhat confusing usage tips:

@DanInci thoughts?

```bash
ob run benchmark --out foo -b ../clustering_example/Clustering_conda.yml
Benchmark YAML file integrity check passed.
Running benchmark...
usage: ob [-h] [--dry-run] [--profile PROFILE] [--workflow-profile WORKFLOW_PROFILE] [--cache [RULE ...]] [--snakefile FILE] [--cores N] [--jobs N]
          [--local-cores N] [--resources NAME=INT [NAME=INT ...]] [--set-threads RULE=THREADS [RULE=THREADS ...]] [--max-threads MAX_THREADS]
          [--set-resources RULE:RESOURCE=VALUE [RULE:RESOURCE=VALUE ...]] [--set-scatter NAME=SCATTERITEMS [NAME=SCATTERITEMS ...]]
          [--set-resource-scopes RESOURCE=[global|local] [RESOURCE=[global|local] ...]] [--default-resources [NAME=INT ...]]
          [--preemptible-rules [PREEMPTIBLE_RULES ...]] [--preemptible-retries PREEMPTIBLE_RETRIES] [--configfile FILE [FILE ...]]
          [--config [KEY=VALUE ...]] [--replace-workflow-config] [--envvars VARNAME [VARNAME ...]] [--directory DIR] [--touch] [--keep-going]
          [--rerun-triggers {code,input,mtime,params,software-env} [{code,input,mtime,params,software-env} ...]] [--force]
          [--executor {slurm,slurm-jobstep,local,dryrun,touch}] [--forceall] [--forcerun [TARGET ...]]
          [--consider-ancient RULE=INPUTITEMS [RULE=INPUTITEMS ...]] [--prioritize TARGET [TARGET ...]] [--batch RULE=BATCH/BATCHES]
          [--until TARGET [TARGET ...]] [--omit-from TARGET [TARGET ...]] [--rerun-incomplete] [--shadow-prefix DIR]
          [--strict-dag-evaluation {cyclic-graph,functions,periodic-wildcards} [{cyclic-graph,functions,periodic-wildcards} ...]]
          [--scheduler [{ilp,greedy}]] [--scheduler-ilp-solver {PULP_CBC_CMD}] [--conda-base-path CONDA_BASE_PATH] [--no-subworkflows]
[...]

PR #192, opened 2025-11-11
bug: Fix confusing error message when wrong flags are passed

DanInci:
## Ticket

Related to https://github.com/omnibenchmark/omnibenchmark/issues/189

## Description

Since allowing extra arguments to be passed by `ob run` commands directly to snakemake, some error messages are confusing. Check ticket above.

In this PR, I'm trying to catch snakemake argument parsing messages and append some useful context for the user:

```
================================================================================
HINT: The error above is from Snakemake's argument parser.
If you intended to use an ob-specific flag, run 'ob run benchmark --help' for more info.

All other flags are passed to Snakemake.
================================================================================
```

## Checklist:

### DIFF
--- a/tests/workflow/test_snakemake_engine.py
+++ b/tests/workflow/test_snakemake_engine.py
@@ -1,0 +1,155 @@
+"""Tests for SnakemakeEngine coverage gaps."""
+
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from omnibenchmark.workflow.snakemake.snakemake import (
+    SnakemakeEngine,
+    _print_ob_flags_help,
+)
+
+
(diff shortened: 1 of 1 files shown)


btraven00 (approved):

imallona (changes requested):
What if `-- --threads` (snmk) and `--cores` (ob) are passed together? and what if not `--dry-run` is used, but `-- --dry-run` is? Is this system aware of the snmk/ob synonyms? Which one takes precedence?

DanInci:
The snakemake specific arguments are constructed in `snakemake.py::_prepare_argv`
Snakemake uses `argparse` to parse arguments, and we pass the snakemake specific arguments after passing the known ob arguments. Given this, for the precedence, the last provided value will be used.

Issue #176, opened 2025-10-09
AmbiguousRuleException in case output file paths are confusable


DanInci:
There's a problem in case the output path for different modules (from different stages) is the same, despite having unique ids:
Try running the following [vizbench.yaml](https://github.com/user-attachments/files/22796823/vizbench.yaml).

Workaround will be renaming `{dataset}.ad` into unique `{dataset}_raw.ad`, `{dataset}_simulated.ad` and `{dataset}_normalized.ad`

Stack trace:
```
AmbiguousRuleException:
Rules normalize_log1pCP10k_.e20ecb3e2b36a2aac4acc23942f4db971db9ff1df7da16379f2827a9ae1c1f2d and simulate_scdesign3_.06aea85bb999c0398a2a7c1a08e86d1b422d4e47769d50f09d32814fd33901c2 are ambiguous for the file out/rawdata/mouse_pancreas/.fc73cbc33384e8af805caeef63502009dc5d075fed448a55aa108e5cf0fc651a/simulate/scdesign3/.06aea85bb999c0398a2a7c1a08e86d1b422d4e47769d50f09d32814fd33901c2/normalize/log1pCP10k/.e20ecb3e2b36a2aac4acc23942f4db971db9ff1df7da16379f2827a9ae1c1f2d/mouse_pancreas.ad.
Consider starting rule output with a unique prefix, constrain your wildcards, or use the ruleorder directive.
Wildcards:
        normalize_log1pCP10k_.e20ecb3e2b36a2aac4acc23942f4db971db9ff1df7da16379f2827a9ae1c1f2d: dataset=mouse_pancreas,post=normalize/log1pCP10k/.e20ecb3e2b36a2aac4acc23942f4db971db9ff1df7da16379f2827a9ae1c1f2d,pre=out/rawdata/mouse_pancreas/.fc73cbc33384e8af805caeef63502009dc5d075fed448a55aa108e5cf0fc651a/simulate/scdesign3/.06aea85bb999c0398a2a7c1a08e86d1b422d4e47769d50f09d32814fd33901c2
[...]

PR #194, opened 2025-11-11
bug: Fix scenarios where AmbiguousRuleException appears (Closes: #176)

DanInci:
## Ticket

Solves: https://github.com/omnibenchmark/omnibenchmark/issues/176

## Checklist:


btraven00 (approved):

btraven00:
Arent' we just trying to catch an enforcement of a snakemake rule in a way that makes sense to the user, rather than letting snakemake invocation fail at runtime? If so, I'd say we're not adding any new feature, just adding clarity to the logical application of snakemake internal consistency rules.

DanInci:
One could redesign your example as this, and it will work

```mermaid
graph LR
    A["Stage: generates_two_things"] --> B["Stage: ingests_two_things → data1.matrix"]
    C["Stage: generates_three_things"] --> D["Stage: ingests_three_things → data2.matrix"]
    B --> C
    D --> E["data1.matrix + data2.matrix → Stage: ingests_data_matrices"]
```

DanInci:
@imallona Previous review / comments are no longer relevant.
Managed to fix the issue, rather than the symptom.

tests/model/test_model.py
```diff
@@ -536,4 +498,6 @@ def test_full_benchmark_workflow(self, tmp_path):
         assert benchmark.id == "integration_test"
 
         # Validate software environments
-        benchmark.validate_software_environments()
+        errors = []
+        benchmark._validate_software_environments(errors)
```
btraven00: Side note, but this works nicely with moving this validation to a specific validation right?

thinking on the next CLI step:

- 1. validate plan: just the YAML "coherence" and syntatic validity
- 2. validate environments: check that this benchmark can be run (has the required environments; implicitely called when running
- 3. validate module: validates module metadata (needs to clone)
- 4. validate outputs: get one output folder and validate the artifact rules

the work you're doing here is still conceptually part of 1, correct? (i.e. the environment "rules" and declaration match whatever is specified in the YAML; concordance of benchmark-level backend with per stage declarations).

omnibenchmark/model/validation.py
```diff
@@ -165,75 +165,66 @@ def _validate_environment_references(self, errors: List[str]) -> None:
...
-            elif self.software_backend == SoftwareBackendEnum.envmodules:  # type: ignore
-                if not env.envmodule:  # type: ignore
-                    errors.append(
-                        f"Cannot use envmodules backend, no envmodule configuration for environment '{env.id}'"  # type: ignore
-                    )
+        elif software_backend == SoftwareBackendEnum.conda:
+            environment = software.conda
 
-        # Check for unused environments
-        unused_envs = env_ids - used_env_ids  # type: ignore
```
btraven00: are you removing the warning for unused to simplify or?

imho it helps keeping a clean YAML and removing potentially unused blocks.

omnibenchmark/model/validation.py
```diff
@@ -165,75 +165,66 @@ def _validate_environment_references(self, errors: List[str]) -> None:
...
 
-    # Note: The old validate_structure method has been split into:
-    # - validate_model_structure() for pure model validation (above)
-    # - validate_execution_context() in the Benchmark class for path validation
+        if BenchmarkValidator.is_url(
+            environment
+        ) or BenchmarkValidator.is_absolute_path(environment):
+            environment_path = environment
+        elif software_backend == SoftwareBackendEnum.envmodules:
+            environment_path = environment
```
btraven00: I think if merging this first (that we should) we will not do this other fix:

https://github.com/omnibenchmark/omnibenchmark/pull/219/commits/b28fa746310e416e8b59bcda3f5b6e2c2d12452b

Issue #173, opened 2025-09-01
ob writes files in /tmp instead of the temporary directory


DanInci:
`ob run module` uses the `/tmp` directory instead of `TMPDIR` environment variable.

**Impact**
In multi-tenant setups, this can lead to concurrent write conflicts and unexpected behavior, since different users/processes share the same /tmp namespace.

**Proposed Solution**
Update the code to respect `TMPDIR` (falling back to `/tmp` only if it is unset), ensuring temporary file isolation across users.

**Steps to Reproduce**
- Run ob run module as two different users at the same time.
- Observe that both write into /tmp, causing collisions.

```
 reason: Missing output files: /tmp/rawdata/mouse_pancreas/.fc73cbc33384e8af805caeef63502009dc5d075fed448a55aa108e5cf0fc651a/mouse_pancreas.sce.rds
    resources: tmpdir=/home/user/tmp

Traceback (most recent call last):
  File "/home/user/src/vizbench-benchmark/.snakemake/scripts/tmpmp1os0fk.run_module.py", line 55, in <module>
    manager.store(parameters)
  File "/home/user/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/omnibenchmark/benchmark/symlinks.py", line 117, in store
    with self.lock:
  File "/home/user/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/filelock/_api.py", line 376, in __enter__
    self.acquire()
  File "/home/user/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/filelock/_api.py", line 332, in acquire
    self._acquire()
  File "/home/user/miniconda3/envs/omnibenchmark/lib/python3.12/site-packages/filelock/_unix.py", line 42, in _acquire
    fd = os.open(self.lock_file, open_flags, self._context.mode)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/tmp/rawdata/mouse_pancreas/storage.lock'
```

imallona:
Is this fixed in https://github.com/omnibenchmark/omnibenchmark/blob/v0.3.1/CHANGELOG.md @DanInci ?

imallona:
[...]

PR #198, opened 2025-11-13
bug: fix hardcoded tmp

btraven00:
## Ticket

#173 

## Description

Avoid hardcoding /tmp folder

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import tempfile
 from itertools import chain
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


btraven00:
A port of this has been merged into the 0.3.x branch, and released 0.3.2
Please review at your convenience so that we can have dev up to date.

DanInci (approved):

omnibenchmark/cli/run.py
```diff
@@ -279,7 +280,7 @@ def run_module(
     logger.info("Running module on a local dataset.")
 
     try:
-        b = BenchmarkExecution(Path(benchmark), Path("/tmp"))
+        b = BenchmarkExecution(Path(benchmark), Path(tempfile.mkdtemp()))
```
btraven00: @imallona you mentioned this is not what you want because this should use a temporary file for each run and not using the env variable.

please have a look at the documentation https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp and let me know if you'd like any changes. thanks!

PR #215, opened 2025-11-21
docs: update readme

btraven00:
merge readme from main; fixes formatting issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,10 @@
 ## omnibenchmark
 
-<p align="center">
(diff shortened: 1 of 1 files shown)


btraven00:
if you mean this: 
https://github.com/omnibenchmark/omnibenchmark/pull/185/files#diff-7618337a88ea47f9b778a21f0e17d43eab2aba4880c6be2d95a7bf95e4b8b2dfR14

I think it might not be "broken". the link you paste is built from main isn't it?

afaik you only removed it from the navigation bar.

I don't fully understand the reason for the removal (I don't know what "it doesn't materialize as a file with current ob" is supposed to refer to), but we can remove the reference from the readme if you feel it's needed.

PR #217, opened 2025-11-24
feat: add validation engine

btraven00:
## Ticket

## Description

## Checklist:

## Remarks for the reviewer:

Just the validation engine. Check against the approved design document.

### DIFF
--- a/docs/design/002-module-artifact-validation.md
+++ b/docs/design/002-module-artifact-validation.md
@@ -5,11 +5,11 @@
 
 **Authors**: ben, ...
 **Date**: 2025-07-08
-**Status**: Draft
-**Version**: 0.1
+**Status**: Approved
+**Version**: 0.1.0
 **Supersedes**: N/A
-**Reviewed-by**: TBD
-**Related Issues**: TBD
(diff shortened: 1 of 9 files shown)


btraven00:
Closing this since it's stale for many months.

The merging window was before freezing the CLI and defining the validation commands. In an omnibenchmark meeting with @DanInci and @imallona  (to which I was referring above) it was agreed that we didn't want incremental merges, and that any functionality like this engine was provided should be backed up with end-to-end CLI functionality.

I've decided to change strategy and try a more bottom-up approach to validation. For someone curious or reading this: this mean that we trust benchmarkers to know their domain better than the ob authors. This means that validators can be defined for each stage I/O contract as arbitrary entrypoints, in a separate repo or integrated with the benchmark itself. This is also linked to the question of providing "valid" inputs (samples, test slices) and the question of "boilerplate" code (since it deals with readers and writers that can be reused across modules sharing a stage.

However, a few things need to be landed first to make this experience smooth.

For the time being, I think it's more sensible to experiment with benchmark-specific validation scripts. Then, learning from those experiences, we can incorporate whatever patterns emerge into the tool itself.

PR #243, opened 2025-12-04
fix: appropriate passing of --task-timeout to rules

DanInci:
## Description

In the scope of this PR there is:
- appropriate interface, docstrings and optional task_timeout int parameter passing
- optional passing of --task-timeout to run module command
- by default --task-timeout is set to 4h
- can override --task-timeout to none, such that there's no timeout

## Checklist:


btraven00 (approved):
lgtm!

I think a minimal unittest (for args on the benchmark/entrypoint) will be enough to pass the global coverage.

as a side note, the forced repetition of the same argument in many places makes me think we could use a configuration pattern around the next time we try to refactor execution code.

PR #247, opened 2025-12-05
feat: per user tmpfile

btraven00:
using specific per-user tmp folders

### DIFF
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -1,8 +1,10 @@
 """cli commands related to benchmark/module execution and start"""
 
+import getpass
 import os
 import sys
 import tempfile
+from datetime import datetime
 from itertools import chain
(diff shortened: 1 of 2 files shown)


omnibenchmark/cli/run.py
```diff
@@ -23,6 +25,19 @@
...
+    """
+    Create a temporary directory with a prefix containing username and timestamp.
+
+    Returns:
+        str: Path to the created temporary directory
+    """
+    username = getpass.getuser()
+    timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+    prefix = f"ob-run-{username}-{timestamp}-"
+    return tempfile.mkdtemp(prefix=prefix)
```
DanInci: Why is this needed, why `tempfile.mkdtemp()` does not suffice ?

PR #255, opened 2025-12-08
feat: cite and validate commands (Closes: #229)

btraven00:

PR #256, opened 2025-12-09
chore: Integration branch for  0.4.0 pre1

btraven00:
triggering a build to test dev + fix templates (0.4.0pre1) + validate
adding code from #254 and #255


btraven00:
I've reviewed the doc changes, and all conflicts should be merged with `-X=theirs` when bringing onto main.

PR #276, opened 2026-01-20
bug: deleting intermediate results should trigger re-running of the benchmark

### LINKED ISSUE #277: deleting intermediate results does not trigger re-running of the benchmark
When deleting an intermediate result, subsequent runs should re-generate the missing files.
We need a regression test for this case.

DanInci:
- Fix dirty rerun for snakemake. In case files are missing or checksum has changed, automatically rerun necessary rules
- Fix output path prefix for metric collector outputs

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -338,7 +338,7 @@ metric_collectors:
       - metrics.json
     outputs:
       - id: plotting.html
-        path: "{input}/{name}/plotting_report.html"
+        path: "{name}/plotting_report.html"
 ```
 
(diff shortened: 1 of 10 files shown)


docs/src/tutorial.md
```diff
@@ -338,7 +338,7 @@ metric_collectors:
       - metrics.json
     outputs:
       - id: plotting.html
-        path: "{input}/{name}/plotting_report.html"
+        path: "{name}/plotting_report.html"
```
btraven00: 👍 
reminder, we should also update clustering_example after this change.

not totally clean yet (since we allow slashes here by passing the module name explicitly, which we don't in modules) but way less confusing.

btraven00 (approved):

PR #278, opened 2026-01-20
tests: add failing regression test

btraven00:
## Ticket

Regression test for #277


DanInci:
@btraven00 make it fail first maybe, and then rebase once #276 is merged

DanInci (approved):

PR #291, opened 2026-02-20
docs: add extensions to YAML spec

btraven00:
This PR gathers extensions to the YAML spec that I added while working on the exploratory explicit snakefile feature branch.

This document gathers different, unrelated extension proposals that are made possible by the explicit snakefile:

# 1.   Named Entrypoints (§3.5)

- Modules can now declare multiple entry scripts in their omnibenchmark.yaml via an entrypoints: dict
- Default behavior preserved for backward compatibility (uses "default" key)
- Enables reusing a single repository for multiple pipeline steps without code duplication

# 2. Gather Stages (Section 7)

Problem solved: Previously, metric_collectors could aggregate results but were a dead-end (outputs couldn’t be chained). Gather stages make reduce operations first-class and composable.

# 3. Output Path Templates (§7.7)

Rich templating system for output paths with variables:

* `{label}`: module ID from ancestor providing that label
* `{params.key}`: parameter values
* `{module.id}`, `{module.stage}`, `{module.parent.id}`: structural attributes
* `{dataset}`: backward-compat alias (planned for deprecation)

# 4. Resource Allocation (Section 8)

Explicit resource management for scheduling:

```
resources:
  cores: <integer>
  mem_mb: <integer>
  disk_mb: <integer>
  runtime: <integer>
``` 

* Can be declared at stage, module, or metric collector level
* Module-level overrides stage-level (most-specific-wins)
* Default: cores=2 when unspecified
* Maps directly to Snakemake resources: and threads: directives

The document is presented here for a cohesive reference, but we can discuss each of these points in a different issue.

### DIFF
--- a/docs/design/004-yaml-specification.md
+++ b/docs/design/004-yaml-specification.md
@@ -1,15 +1,15 @@
-# 004: Omnibenchmark YAML Specification (v0.3)
+# 004: Omnibenchmark YAML Specification (v0.1)
 
-[![Status: Accepted](https://img.shields.io/badge/Status-Accepted-green.svg)](https://github.com/omnibenchmark/docs/design)
-[![Version: 0.3](https://img.shields.io/badge/Version-0.3-blue.svg)](https://github.com/omnibenchmark/docs/design)
+[![Status: Under Review](https://img.shields.io/badge/Status-UnderReview-yellow.svg)](https://github.com/omnibenchmark/docs/design)
+[![Version: 0.1](https://img.shields.io/badge/Version-0.3-blue.svg)](https://github.com/omnibenchmark/docs/design)
 
-**Authors**: ben,
+**Authors**: ben
 **Date**: 2025-01-20
 **Status**: Under Review
 **Version**: 0.1
 **Supersedes**: N/A
-**Reviewed-by**: TBD
-**Related Issues**: TBD
+**Reviewed-by**: daninci
+**Related Issues**: #283
 
 ## Changes
 
@@ -189,7 +189,6 @@ Modules MUST NOT declare these in their YAML configuration.
 #### Parameter Values
 
 - **String**: Passed directly (`evaluate: "1+1"` → `--evaluate 1+1`)
-- **List**: Joined with commas (`items: [a, b, c]` → `--items a,b,c`)
 
 #### Wildcard-Based Resolution
 
--- a/docs/design/004-yaml-specification.md
+++ b/docs/design/004-yaml-specification.md
@@ -1,21 +1,24 @@
-# 004: Omnibenchmark YAML Specification (v0.1)
+# 004: Omnibenchmark YAML Specification (v0.5)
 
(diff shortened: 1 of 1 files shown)


docs/design/004-yaml-specification.md
DanInci: Mention here that `host` is a reserved keyword for a software_environment. While the other can be freely defined names.

DanInci: We need to present a list of fully supported wildcard names.
We only support `{dataset}`, `{params}`, `{stage}` and `{module}` as of now. And some of these can only be used as internal wildcards, and definitely not exposed. We need to rethink the system.

docs/design/004-yaml-specification.md
```diff
@@ -445,3 +495,443 @@ Backends must support:
...
+Currently, all map stages that share a dependency chain are fully expanded into a cartesian product: every dataset × every method × every metric. The only escape hatch is `exclude:` on individual modules.
+
+`provides` labels open a path to a more explicit contract: a downstream stage could declare `needs: {method: fast}` to depend only on stages that provide the `method` label *and* match a tag. This would allow selective pairing — "metric M only applies to fast methods" — without enumerating exclusions. The `provides`/`needs` pair would replace the implicit "anything upstream that shares an output ID" wiring with an explicit, typed contract, reducing the cartesian product to only the declared combinations. This extension is a natural continuation of the label system already in place.
+
+---
+
+## 8. Resource Allocation (v0.4+)
+
+> **Status:** Implemented.
+
```
DanInci: Is it implemented tho?

DanInci (changes requested):

docs/design/004-yaml-specification.md
btraven00: yes

docs/design/004-yaml-specification.md
```diff
@@ -445,3 +495,443 @@ Backends must support:
...
+Gather stages generalise the reduce pattern: any stage can gather all outputs from any set of provider stages, and gather stage outputs participate in the DAG like regular stage outputs. This makes reduce composable with further map or reduce steps.
+
+**Why `provides` is needed.** A stage may declare multiple outputs (e.g. `data.raw` and `data.metadata`). When a downstream stage gathers from that stage, it must name *which* output to collect — referencing the stage ID alone is ambiguous. `provides` is the mechanism that binds a human-readable label to a specific output ID, and that label is what gather consumers reference. This also decouples the gather contract from internal output naming.
```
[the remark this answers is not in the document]
btraven00: if I see the argument right, you're proposing to pause on provides, and just implement gather as a explicit list of stage + output, correct?

imallona (approved):

btraven00:
Section 8 (resources) has been merged to main after landing #298

DanInci (approved):

btraven00:
Leaving this PR open for historical reasons, but it has been landed incrementally. Need to review which parts are not yet merged (the gather, mostly) and split them apart for future work in 0.7.0 or 0.8.0

btraven00:
closing, this has been landed

PR #303, opened 2026-03-10
chore: update CITATION.cff, add action to auto-update

btraven00:
Semantically, we're decoupling the authorship of the paper and the software here.
Using an action I developed during the snakemake hackathon, further testing might be needed.

### DIFF
--- a/.github/workflows/cffbump.yaml
+++ b/.github/workflows/cffbump.yaml
@@ -1,0 +1,23 @@
+name: Update CITATION.cff
+
+on:
+  push:
+    branches:
+      - main
+    tags:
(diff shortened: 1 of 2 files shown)


DanInci (approved):

imallona (approved):
It would be nice to have a "good practices" or "authorship thoughts" section on the docs on how to handle the three sides of benchmarking modules/yamls: original software developers (and their papers; imagine a module wrapping Seurat); module/yamls developers (people developing the omni-seurat-bench; module/yaml associated papers and their authors (the omni-seurat-bench paper).

Issue #313, opened 2026-04-16
parameter expansion: raise warning for disjoint keys


btraven00:
One common gotcha, given the flexibility to pass parameter sets with list-based expansion, is that the user might enter this:

```
parameters:
          - selection_type: ["seurat_vst", "scrapper_modelGeneVariances"]
          - number_selected: 2000
```

When meaning this instead:

```
parameters:
          - selection_type: ["seurat_vst", "scrapper_modelGeneVariances"]
            number_selected: 2000
```

The simplest solution might be to raise a prominent warning in the first case, assuming the heuristic of disjoint set of keys.

We should prompt the user a hint like "WARN: you specified a list of parameter with ... that will expand into ..., but no common parameters found. are you sure you don't mean ... instead?".

btraven00:
this was done. if this gets annoying as a warning, we can silence it with a flag/config entry.

PR #320, opened 2026-04-28
fix: raise warning for disjoint keys

btraven00:
## Ticket

#313

## Description

It's unlikely that the user really intends to pass disjoint parameters to a method. Raising a hint for a common gotcha.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This document records all notable changes to `omnibenchmark`.
 This project adheres to [Semantic Versioning](https://semver.org/) and [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
  
+## [0.5.1](UNRELEASED) ()
+
+- fix: raise warning for disjoint keys (Closes: #313)
+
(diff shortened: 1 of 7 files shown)


btraven00:
Screenshot_2026-04-28_12-06-19

markrobinsonuzh:
did a test this morning .. should this construction of params throw a warning?
```
        parameters:
          - filter_type: ["manual", "scrapper-auto"]
          - filter_type: ["manual", "scrapper-auto"]
            this: ["a", "b"]
```
(it does not)

but this construction does throw a warning:

```
parameters:
          - filter_type: ["manual", "scrapper-auto"]
          - this: ["a", "b"]
``

I'm using `OmniBenchmark CLI, version 0.5.0.post4+ga01f1165d.d20260429` here.

btraven00:
Good question. I think parameter expansion might not be caught in the tests.

@markrobinsonuzh do you think we should expand the warning to the first case too? I see it might be just a slipped typo that changes semantics.

Worst case, if someone has a good case to use that, they can pass a flag to silence the warning...

markrobinsonuzh:
Yes, I think we should expand the warning for that case above.

markrobinsonuzh:
Looks good in my test!

markrobinsonuzh (approved):

Issue #321, opened 2026-04-28
bug: misleading error when a named entrypoint is missing


btraven00:
`omnibenchmark/backend/resolver.py`

## Problem

 When a benchmark requests repository.entrypoint: <name> and that key isn't in the module's omnibenchmark.yaml, the resolver raises:

  ▎ Failed to dereference entrypoint for module ''. Expected omnibenchmark.yaml or config.cfg in

  …but the file is there — only the requested key is missing. _read_entrypoint (≈L660–710) correctly logs the real cause ("Entrypoint 'X' not found. Available: …"), but the caller (≈L278–284) collapses every None return into the same generic message, contradicting the log.

##  Reproduce

  Module's omnibenchmark.yaml:

```
  entrypoints:
    default: run.py

  Benchmark YAML:
  repository:
    url: ../path/to/module
    entrypoint: pca-scanpy   # not defined above
    commit: main
```

  Run `ob run <benchmark.yaml>` → generic "Expected omnibenchmark.yaml or config.cfg" error.

##  Expected

  Entrypoint 'pca-scanpy' not found in module 'five-pca/pca-scanpy'.
  Available entrypoints: default

## Fix

  Have _read_entrypoint return (or raise) a typed reason; surface it in the caller's error instead of the generic message.

PR #322, opened 2026-04-28
fix: clearer error on missing entrypoint

btraven00:
## Ticket

#321

## Description

When a benchmark requests repository.entrypoint: and that key isn't in the module's omnibenchmark.yaml, the resolver raises a misleading error.

## Checklist:

## Testing

This can be tested manually with an omnibenchmark.yaml for a module where you request a missing entrypoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 This document records all notable changes to `omnibenchmark`.
 This project adheres to [Semantic Versioning](https://semver.org/) and [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- 
+
 ## [0.5.1](UNRELEASED) ()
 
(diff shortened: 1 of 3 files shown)


DanInci:
Do we mark at benchmark yaml validation the fact that the `config.cfg` usage is deprecated? 
Otherwise this might look like a surprise error.

DanInci (approved):

btraven00:
This is the notice that we currently give if there's an old-style `config.cfg` file and no omnibenchmark.yaml:

```
Validating module at: /home/b/lab/omni-scrna/split-stages-plan-modules/4-select
Using legacy config.cfg file. Please migrate to omnibenchmark.yaml
```

(in retrospective I regret we didn't choose `omni-module.yaml` but I feel that ship has sailed for now :ship:

PR #323, opened 2026-04-28
feat: resolve module name correctly

btraven00:
## Ticket

...

## Description

Fix the wrong behavior by which dataset name was propagated as name. name should be the module's id. ensure other variables are working, and document them properly in the user-facing site.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/) and [Conventi
 
 ## [0.6.0](UNRELEASED)
 
-- ...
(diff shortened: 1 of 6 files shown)


DanInci (approved):
I am actually surprised all these keywords are supported with such minimal changes !
Looks really great. If there aren't any regressions in the tests, this seems ready to go.

btraven00:
@DanInci thanks for reviewing! I've squashed the commits and changed the milestone and CHANGELOG entry. I feel now this is better scheduled for the 0.6.0 milestone. We realized later that introducing naming flexibility breaks modules and metric collector - I've worked on #329 on top of this. Still worth to land and test properly this commit first for the 0.6 release.

btraven00:
rebased to current main

btraven00:
I think we need to add tests in which we bump explicitly the api_version to 0.6, because otherwise we'll be masking the default 0.4 behavior for backwards compat.

PR #329, opened 2026-05-05
feat: named outputs

btraven00:
## Ticket

## Description

## Checklist:

## Remarks for the reviewer:


docs/design/009-named-outputs.md
```diff
@@ -0,0 +1,224 @@
...
+parallel access for downstream readers when entries are stored uncompressed.
+
+### 3.2 Named output contract (API v0.6)
+
+For benchmarks declaring `api_version: 0.6.0`, the runtime passes one
+`--output` flag per declared stage output, in addition to the existing
+`--output_dir` and `--name` flags (which remain for backward compatibility
+through v0.7, dropping in v0.8).
+
+**Multi-output stage:**
```
atchox, line 87: My first comment is really a question for my own understanding. Why would you not do something like 

```
<entrypoint> \
    --output_dir $OUTPUT_DIR \
    --name <module_id> \
    --rawdata.h5ad /abs/path/dataset1.h5ad \
    --rawdata.clusters_truth /abs/path/dataset1.clusters_truth.tsv \
    --input_data /abs/path/to/input.h5ad \
    [param flags]
```

for the outputs? Why do you need the separate --output arg? Do the namespaces collide..? Looking at the shell directives in the explicit Snakefile generation, this seems feasible if the namespaces don't collide. At the end, the approach you have proposed here will disentangle file paths from the modules, which is a win.

With this in place, I think there is a chance to go one step further for module reusability across benchmarks. I will mix stages and modules here because I am concerned about inputs and outputs which is set at the stage level. Consider two benchmarks, B1 and B2, identical in spirit, each with two stages. Let the first stage, both in B1 and B2 be D. Let the second stage in B1 and B2 be M1 and M2 respectively. So you have the flow,

B1: D -> M1 and B2: D -> M2

Both M1 and M2 expect a dataset. The author of M1 likes more explicit arg names and chooses `dataset.path` as the input while the author of M2 likes to keep things succinct and prefers `d.p` for the same argument. How would module D fit into both benchmarks? Currently, the argument names as seen by the modules are the same as seen by the benchmarks. They are used to resolve the input path of child stages from the output path of parent ones (I presume) during Snakefile generation. By doing this, you set a restriction on the output arg names of D and input arg names of M1 and M2. This could be solved with another namespace. Consider the syntax below for the two plans.

```yaml
# Benchmark B1
stages:
  - id: D
    outputs:
      - id: B1_dataset_label
        internal: illuminatiIsReal
        path: "{dataset}.h5ad"

  - id: M1
    inputs:
      - id: B1_dataset_label
        internal: dataset.path
```

```yaml
# Benchmark B2
stages:
  - id: D
    outputs:
      - id: B2_dataset_label
        internal: illuminatiIsReal
        path: "{dataset}.h5ad"

  - id: M2
    inputs:
      - id: B2_dataset_label
        internal: d.p
```

You would resolve the paths in M1 and M2 with B1_dataset_label and B2_dataset_label respectively. It also moves the onus to the benchmarker to keep that namespace clean. Module authors can name their args however they like as long as they adhere to some specifications (validators!!!).

Now...having typed out all this, I think I just revisited the discussion on `provides`, albeit with a different syntax. Still, keeping this for posterity.

_[Jumping back on point 1 after writing discussing point 2, I see how --output avoids the possible collision. However, I don't like how a module author would have to resolve that...maybe a prefix to differentiate between input and output args?]_

Thanks for reading!

docs/src/howto.md
```diff
@@ -424,6 +424,97 @@ stages:
...
+        name: "Method 2"
+        ...
+    outputs:
+      - id: methods.result
+        path: "{name}_result.txt"    # → M1_result.txt / M2_result.txt  (own module ID)
+        # or equivalently:
+        # path: "{module.id}_result.txt"
+```
+
+The `--name` CLI argument passed to each module script always receives the current module's own ID, matching the `{name}` template variable.
```
atchox, line 540: :)

btraven00: > Why do you need the separate --output arg? Do the namespaces collide..? Looking at the shell directives in the explicit Snakefile generation, this seems feasible if the namespaces don't collide.

The only problem I see with the option you suggest is that parsing is more cumbersome on the receiving side. If you know you have to parse one ore more `--output` flags you can, otherwise you have to loop over all the flags. TBH I would recommend module authors to be very strict in validating unknown flags (to catch flag drift, typos etc)

docs/design/009-named-outputs.md
btraven00: I think you're hitting the nail in the head here; this is yet another instance of coupling the particulars of one benchmark to the modules implementation.

Why named outputs you ask?

Because it tighten the spec. The contract just becomes about the semantics (an "embedding" is anything I know how to unpack to a 2D array of cell_ids x components, with dtype=float). i/o becomes a pluggable reader-writer convo that can be provided centrally. replacing a tsv with a h5 writer becomes just picking a serialization "flavor" (I'm being handwavy about the details: this assumes there's a way in which information flows from the benchmark to the modules. other systems solve this by imposing a structured api common to every benchmark; here we additionally have to cross the language barrier).

And the only surface they need to agree upon is a label, "embedding" (plus the agreed semantics, that can be just documented in a .md, no need to go formal on this). This assumes there's a registry of  modules where we avoid clashes.

Incidentally, [snakemake might support typed outputs](https://github.com/snakemake/snakemake/pull/4035#discussion_r2913404691) in checkpoints, which we could use in the future to validate if two modules can be connected in a pipeline.

PR #293, opened 2026-02-20
docs: design document for telemetry functionality

btraven00:
Design document for telemetry functionality

PR #333, opened 2026-05-07
feat: add OpenTelemetry support

btraven00:
Ports OpenTelemetry instrumentation from the execution-shell-telemetry branch (squash of 13d2700, dc9daf0, afe2829) onto current main.

Adds:
- omnibenchmark/telemetry/ — pure-Python OTLP-shaped emitter, span builder, and event types (no external opentelemetry-sdk dependency).
- scripts/telemetry-relay.py — relay for forwarding JSONL telemetry.
- --telemetry [json] and --telemetry-output PATH options on 'ob run'. Streaming JSON to stdout disables the Rich progress UI; --telemetry-output writes to a file and keeps Rich active.
- Per-rule log capture wired through the existing log: directive in the generated Snakefile and surfaced via _read_rule_log.

Drops the older snakemake_gen.py touches: the equivalent log/tee infrastructure already lives in backend/snakemake.py on main. Drops the older cache.py changes: superseded by main's refactored ref resolution.

Refs: #293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/) and [Conventi
 
 ## [0.6.0](UNRELEASED)
 
+- feat: add opentelemetry support
 - feat: output path templates now support `{name}`, `{module.name}`, `{module.stage}`, and `{params.KEY}`; `{name}` always resolves to the current module's own ID (never inherited)
 
 ## [0.5.3](https://github.com/omnibenchmark/omnibenchmark/releases/tag/v0.5.3) (Jun 15th 2026)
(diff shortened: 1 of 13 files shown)


btraven00:
To test this, https://github.com/omnibenchmark/obmon can be used

markrobinsonuzh:
Looks really good to me. I tested it and it's easy to run. I would only suggest:
- a little extra reminder in the docs about where each things run (e.g., if running a b/m on a server, that `obmon` runs locally
- if there is a way to simplify the interface from the telemetry output browser (for me, it required several steps; see below)

markrobinsonuzh:
the process to find the log is not that easy (for me, at least):
- make sure you are on the 'Structured' tab (not 'Traces' because the browser is very slow to render there)
- in the list, find one listed with error (`structuredlogs?logLevel=error` in link might help ..)
- click on FAILED
- click on 'SpanId e37f74d76f1d77d6f5e36f74' (not super obvious that this is the place to get it!)
- scroll down, find 'stderr' and click on the link under the 'Value' column.

markrobinsonuzh (approved):

PR #334, opened 2026-05-08
bug: truncate filename to avoid hitting filesystem limits

btraven00:
eCryptfs might impose similar limits, going with 255 chars now by default but this might be need to be configurable in the future.


btraven00:
@atchox and me tested this and it solves the original bug. ready to merge.

markrobinsonuzh (approved):

Issue #351, opened 2026-06-10
multibyte UTF-8 sequence crashes snakemake logs read with multiple cores in use


btraven00:
There's an intermittent bug surfacing in the CI and reported in a few benchmarks:

Example: https://github.com/omni-scrna/split-stages-plan/actions/runs/27196009666/job/80287698732?pr=39

Its signature is an `UnicodeDecodeError` when decoding stdout from snakemake, with `cores >1`.

```
  File "/home/runner/work/split-stages-plan/split-stages-plan/.pixi/envs/plan/lib/python3.13/site-packages/omnibenchmark/cli/run.py", line 174, in run
    _run_benchmark(
    ~~~~~~~~~~~~~~^
        benchmark_path=benchmark,
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
        snakemake_args=list(snakemake_args),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/runner/work/split-stages-plan/split-stages-plan/.pixi/envs/plan/lib/python3.13/site-packages/omnibenchmark/cli/run.py", line 274, in _run_benchmark
    _run_snakemake(
    ~~~~~~~~~~~~~~^
        out_dir=out_dir_path,
        ^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        extra_snakemake_args=extra_args,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/runner/work/split-stages-plan/split-stages-plan/.pixi/envs/plan/lib/python3.13/site-packages/omnibenchmark/cli/run.py", line 407, in _run_snakemake
    for line in process.stdout:
                ^^^^^^^^^^^^^^
  File "<frozen codecs>", line 325, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 0: invalid continuation byte
```

# Analysis

The error can be traced to how we spawn the snakemake process in ob. Two configurations collide to create the crash: no explicit encoding (can depend on locale), and no explicit error handling in the `Popen`.

Possible fixes:

```
  process = subprocess.Popen(
      cmd,
      stdout=subprocess.PIPE,
      stderr=subprocess.STDOUT,
      text=True,
      bufsize=1,
      encoding="utf-8",   # don't depend on locale
      errors="replace",   # never let bad bytes crash the process
  )
```

PR #352, opened 2026-06-10
fix: avoid crash on logs decode error

btraven00:
## Ticket

#351 

## Description

the subprocess.Popen needs to be made tolerant of decode errors to avoid crashes. This was producing intermittent failures when dealing with invalid byte sequences from interleaved multi-core writes (surfaced when using cores >1)

A second code path is also fixed that re-reads the logs written to files.

## Checklist:


markrobinsonuzh (approved):

markrobinsonuzh:
Yep, looks good!

PR #354, opened 2026-06-15
feat: filtering lineage (provides/requires)

btraven00:
## Ticket

#360

Follow-up on #361 (should be rebased on top of that one when it lands in main; getting 361 as base to avoid reviewing the same code).

## Description

Declares the producer side of the provides/requires lineage gate: 

- stage.provides: list[str] — labels this stage advertises downstream.
- Module.provides: dict[str, str] — explicit per-module label bindings

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -911,14 +911,37 @@ def _substitute_params_in_path(template: str, params) -> str:
     return re.sub(r"\{params\.([^}]+)\}", _replace, template)
 
 
+def _resolve_label_value(label: str, module_provides, module_id: str) -> str:
+    """Resolve a single `Stage.provides` label's value for one node.
+
+    Two-level chain, most specific first:
+      1. ``module.provides[label]`` — explicit benchmark-local binding.
+      2. ``module_id`` — default for the "label = module identity" pattern.
+
+    The parameter-name fallback prototyped earlier is intentionally absent:
+    sourcing a label from a same-named CLI parameter couples the module's CLI
+    contract to the benchmark's routing across repositories. See
+    docs/design/008-filtering.md §3.5.
+    """
+    if module_provides and label in module_provides:
+        return str(module_provides[label])
+    return module_id
+
+
 def _build_template_context(
     stage,
     module_id: str,
     module_name: Optional[str] = None,
     input_node=None,
     params=None,
+    module_provides=None,
 ) -> TemplateContext:
-    """Build a TemplateContext for a node during expansion."""
+    """Build a TemplateContext for a node during expansion.
+
+    `module_provides` is the optional `Module.provides` dict (label → value);
+    it sources the values for the labels this stage advertises via
+    `Stage.provides`, defaulting to the module id.
+    """
     provides: dict[str, str] = {}
     module_attrs: dict[str, str] = {
         "id": module_id,
@@ -928,27 +951,19 @@ def _build_template_context(
 
     stage_provides = getattr(stage, "provides", None)
 
+    # Inherit the upstream lineage labels first, then layer this stage's own.
+    if input_node is not None and input_node.template_context is not None:
+        provides.update(input_node.template_context.provides)
+
+    if stage_provides:
+        for label in stage_provides:
+            provides[label] = _resolve_label_value(label, module_provides, module_id)
+
     if input_node is not None:
-        if input_node.template_context is not None:
-            provides.update(input_node.template_context.provides)
-
-        if stage_provides:
-            for label in stage_provides:
-                if params is not None and label in params:
-                    provides[label] = str(params[label])
-                else:
-                    provides[label] = module_id
-
         module_attrs["parent.id"] = input_node.module_id
         module_attrs["parent.stage"] = input_node.stage_id
     else:
-        if stage_provides:
-            for label in stage_provides:
-                if params is not None and label in params:
-                    provides[label] = str(params[label])
-                else:
-                    provides[label] = module_id
-
+        # Builtin `dataset` label: root identity, propagated downstream.
         if params is not None and "dataset" in params:
             provides.setdefault("dataset", str(params["dataset"]))
         else:
@@ -1326,6 +1341,9 @@ def _generate_explicit_snakefile(
     output_to_nodes = {}
     previous_stage_nodes = []
     dag_errors: list[tuple[str, str, str]] = []
+    # Count pruned combinations by reason, so a silently absent result cell is
+    # observable (see docs/design/008-filtering.md, "No silent absence").
+    prune_counts = {"requires": 0, "exclude": 0}
 
     # stage_id -> ordered output ids. Precomputed once so per-node output lookup
     # is O(1) instead of scanning every stage and doing an O(outputs) pydantic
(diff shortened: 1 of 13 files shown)


omnibenchmark/cli/run.py
```diff
@@ -924,6 +1022,7 @@ def resolve_module_task(cache_key, module, software_env_id):
 
                     if input_node and module.requires:
```
DanInci, line 1464: **Bug: `requires:` is a silent no-op for root-stage modules**

The guard short-circuits when `input_node is None`, which is always the case for modules in the first (data) stage. A module in a data stage that declares `requires: {dataset_size: lg}` will never be pruned — it runs unconditionally on every parameter expansion with no error or warning.

This is probably the most natural place to put a lineage gate on a data-stage variant loader, so the silent skip is likely to surprise users. Consider emitting a parse-time warning (or error) when `module.requires` is set on a module whose parent stage declares no `inputs:`.

omnibenchmark/cli/run.py
```diff
@@ -1071,8 +1171,27 @@ def get_ancestor_nodes(node):
                 if logger.level <= 10:
                     traceback.print_exc()
 
+        # A stage that had modules to expand but produced no nodes was pruned
+        # to empty by a filter (requires / exclude). Without this the empty set
+        # cascades downstream and looks like "nothing happened" — usually a
+        # typo in a label value or an over-broad exclusion.
+        if modules_to_expand and not current_stage_nodes:
+            logger.warning(
+                f"Stage '{stage.id}' produced no nodes: every module/input "
```
DanInci: **Misleading empty-stage warning: fires on bad input reference, not just filter pruning**

The guard `if modules_to_expand and not current_stage_nodes` fires whenever a stage produces zero nodes — including when `_select_input_nodes` returned an empty list because a declared input id doesn't match any output in `output_to_nodes` (a misspelled output id, or a stage whose upstream was skipped for an unrelated reason).

In that case the warning says `"pruned by a filter (requires/exclude)"`, pointing the user at their label configuration when the real cause is a missing/misspelled output reference. The two conditions should be distinguished, e.g. by checking whether `previous_stage_nodes` is empty vs. whether combinations were attempted and individually rejected.

omnibenchmark/model/benchmark.py
```diff
@@ -1297,6 +1328,38 @@ def validate_model_structure_post_init(self) -> "Benchmark":
...
+                            f"requires api_version ≥ 0.6.0 (this benchmark "
+                            f"declares {self.api_version.value})."
+                        )
+
+        # The runtime auto-populates `name` (current module id) and `dataset`
+        # (root dataset identity) on every node; letting a stage advertise
+        # either would silently clobber the user's value and make a downstream
+        # `requires:` gate on an accident of propagation. See 008 §3.5.
+        for stage in self.stages:
+            for label in stage.provides or []:
```
DanInci, line 1392: **Missing validation: `Module.provides` keys are not checked against `Stage.provides`**

A typo in a module's `provides` key (e.g. `daataset_size` instead of `dataset_size`) passes the api_version gate and the reserved-label check here, but is silently ignored at resolution time: `_resolve_label_value` only iterates the labels declared in `Stage.provides`, so a mismatched key is never consulted, and the module falls through to the module_id default.

The downstream `requires: {dataset_size: lg}` then fails to match on the lineage carrying the wrong label value, pruning the module silently — a lost gate with no diagnostic, directly violating the "No silent absence" design goal from 008 §2.

Consider adding a check here: for each module in the stage, warn (or error) if any key in `module.provides` is not in `stage.provides`.

omnibenchmark/cli/run.py
```diff
@@ -1093,9 +1212,20 @@ def get_ancestor_nodes(node):
 
     # Resolve metric collectors (skip in -m module mode)
     if not module_filter:
+        collectors_to_resolve = benchmark.model.get_metric_collectors()
+        if until_stage is not None:
```
DanInci: **Missing case: `_filter_collectors_by_stages` is only called for `--until`, not for organically empty stages**

When a stage is pruned to zero nodes by `requires`/`exclude` filters (rather than by `--until`), `_filter_collectors_by_stages` is never invoked. A metric collector that declares an input from that stage is passed unchanged to `resolve_metric_collectors`, which will find zero matching resolved nodes and either silently produce no output or emit a confusing downstream error.

The same logic that filters collectors for `--until` should also run when any stage has produced zero nodes — or `resolve_metric_collectors` needs to handle the no-nodes case gracefully and emit a clear diagnostic.

omnibenchmark/model/benchmark.py
```diff
@@ -1297,6 +1328,71 @@ def validate_model_structure_post_init(self) -> "Benchmark":
                     if collector.software_environment is None:
                         collector.software_environment = sole_env_id
 
+        # Gate api_version-introduced fields. `Stage.provides` and
+        # `Module.provides` were introduced in 0.6.0 and must not be used by
+        # older specs.
+        if self.api_version < APIVersion.V0_6_0:
```
DanInci: `APIVersion(str, Enum)` has no custom ordering, so this is a plain string comparison, and it's the only place in the model that compares `APIVersion` values with `<`. That happens to work today because every version so far is single-digit-minor, but it breaks the moment a two-digit minor exists. Reproduced directly:

```python
class APIVersion(str, Enum):
    V0_6_0 = "0.6.0"
    V0_10_0 = "0.10.0"

APIVersion.V0_10_0 < APIVersion.V0_6_0  # True (lexicographic "1" < "6")
```

Worth comparing parsed `(major, minor, patch)` tuples instead of relying on string/enum ordering.

DanInci (commented):
Just one finding after the fixes above:

omnibenchmark/model/benchmark.py
btraven00: fixed, reusing `omnibenchmark.versioning`, updated tests

omnibenchmark/cli/run.py
btraven00: done, raising ValueError, tests added

omnibenchmark/cli/run.py
btraven00: Fixed. The stage now counts the (input, params) combinations it actually attempted (combinations_seen) and picks the message from that, rather than reporting every empty stage as filter pruning

omnibenchmark/model/benchmark.py
btraven00: Fixed, as an error instead of a warning, since a provides key that no stage declares can never be queried. added test in `TestLineageProvides::test_module_binds_undeclared_label_rejected`

omnibenchmark/cli/run.py
btraven00: Fixed. _filter_collectors_by_stages now runs unconditionally, and the included set is the stages that actually produced nodes ({n.stage_id for n in resolved_nodes}) rather than the stages --until kept. A stage emptied by requires/exclude isn’t in that set, so its collectors are dropped on exactly the same path as --until truncation

DanInci:
Looks good to me now, but we are missing a CHANGELOG entry for this feature. @btraven00

DanInci (approved):

PR #359, opened 2026-06-25
fix: derive topology edges from declared inputs

btraven00:
Fixes for the mermaid and dot renderers. Also factor out a common functino so that obeditor doesn't need to re-derive the DAG in the javascript code.

The execution DAG connects each node only to its single latest upstream stage so output-path construction stays a linear ancestry. That dropped cross-stage edges whenever an input group referenced outputs from more than one upstream stage.

Derive topology edges from the model's declared inputs instead:
- stage_adjacency / upstream_stage_ids / build_topology_dag in _graph.py
- mermaid generator reads declared inputs (graph arg now unused)
- export_to_dot renders the topology DAG without mutating execution DAG

## Description

## Checklist:

## Remarks for the reviewer:

Test with graphviz and dot code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/) and [Conventi
 
 - feat: add opentelemetry support
 - feat: output path templates now support `{name}`, `{module.name}`, `{module.stage}`, and `{params.KEY}`; `{name}` always resolves to the current module's own ID (never inherited)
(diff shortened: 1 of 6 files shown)


markrobinsonuzh:
Tested it on `omni-scrna` .. it checks out for me .. 

Using 0.5.3
mermaid-0 5 3

Using this branch
mermaid-0 5 3 post11+g11011ff0d

Looks good to go!

markrobinsonuzh (approved):

Issue #289, opened 2026-01-29
support for multi-stage outputs


btraven00:
Currently (0.4.0), it's not possible to make modules in a stage to declare inputs produced in different stages. This is a known limitation, constrained by internal representation in the model.

This forced to separate what it's called `metric collector` as a somewhat privileged gather stage.

Several users are requesting the ability to declare arbitrary multi-stage input collection. This would phase out the need for a special collector, for metrics or other purposes. Under the hood, we might want to have it translated to something along the lines of the gather-scatter pattern  [(see Fig 8a in the snakemake paper)](https://f1000research.com/articles/10-33/v3#f8)

In the meantime, we can document suggested workarounds.

DanInci:
Can be related: https://github.com/omnibenchmark/omnibenchmark/issues/365

DanInci:
### Investigation: two distinct problems hide behind "Could not resolve input `<id>`"

While reproducing this, the opaque runtime warning `Could not resolve input <id>` turned out to have **two separate causes**. Only one of them is the genuine multi-stage-collection limitation this issue tracks. Topologies below are drawn abstractly (edge `X --> Y` means "Y declares an input produced by X"); stage names carry no meaning beyond the graph shape.

#### How the resolver wires inputs

The resolver (`omnibenchmark/cli/run.py`) expands stages and, per stage, calls `_select_input_nodes`, which picks a **single** parent — the *deepest already-expanded* stage producing one of the stage's inputs — then resolves inputs only from that parent's **linear ancestry** (`get_ancestor_nodes`, a node-id prefix walk). Two consequences follow.

#### 1. Declaration order mattered (a bug, not the limitation)

`stages_to_expand` was `benchmark.model.stages` — *declaration order*, not topological order. A stage declared *before* a stage producing one of its inputs couldn't be wired, even when that producer is genuinely upstream on a single chain:

```mermaid
graph LR
  A --> B
  B --> C
  C --> D
  B --> D
  D --> E
  classDef leaf fill:#2da44e,stroke:#1a7f37,color:#fff;
  class E leaf;
```

`E` consumes only `D`'s output. Every producer of `E`'s inputs lies on the one chain `A → B → C → D`, so once stages are expanded in dependency order, `E` resolves — even though it is declared before `C` and `D`. **Not a fundamental limitation.**

#### 2. Diamond joins are genuinely unsupported (this issue)

[...]

PR #361, opened 2026-06-29
feat(cli): add --until STAGE to truncate the resolved DAG

btraven00:
Keeps the named stage plus its transitive ancestors (from declared topology, not YAML declaration order) before module resolution; skips metric collectors whose inputs reference pruned stages; rejects combination with -m/--module.

## Ticket

## Description

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -138,6 +138,18 @@ def format_pydantic_errors(e: PydanticValidationError) -> str:
     default=None,
     help="Write telemetry to file instead of stdout. Allows Rich progress to remain active. Implies --telemetry.",
 )
+@click.option(
+    "--until",
+    "until_stage",
+    default=None,
+    type=str,
+    help=(
+        "Stop the pipeline at and including the named stage. "
+        "Stages declared after the named stage in the benchmark YAML are "
+        "pruned from the resolved DAG, and metric collectors whose inputs "
+        "reference pruned stages are skipped."
+    ),
+)
 @click.argument("snakemake_args", nargs=-1, type=click.UNPROCESSED)
 @click.pass_context
 def run(
@@ -154,6 +166,7 @@ def run(
     module_filter,
     telemetry,
     telemetry_output,
+    until_stage,
     snakemake_args,
 ):
     """Run a benchmark.
@@ -198,6 +211,14 @@ def run(
             f"-m {module_filter}: single execution path, first param expansion only."
         )
 
+    if module_filter and until_stage:
+        log_error_and_quit(
+            logger,
+            "--until and -m/--module cannot be combined: -m already truncates "
+            "the DAG at the target module's stage.",
+        )
+        return
+
     _run_benchmark(
         benchmark_path=benchmark,
         cores=cores,
@@ -211,6 +232,7 @@ def run(
         module_filter=module_filter,
         telemetry=telemetry,
         telemetry_output=telemetry_output,
+        until_stage=until_stage,
         snakemake_args=list(snakemake_args),
     )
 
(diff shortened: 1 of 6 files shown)


DanInci (commented):
(Replaced by line-specific comments below.)

omnibenchmark/cli/run.py
```diff
@@ -138,6 +138,18 @@ def format_pydantic_errors(e: PydanticValidationError) -> str:
     default=None,
     help="Write telemetry to file instead of stdout. Allows Rich progress to remain active. Implies --telemetry.",
 )
+@click.option(
```
DanInci, line 152: This is the only new user-facing flag in the PR, but there's no `docs/src/howto.md` section for it (the sibling capability-gating PR #362 added one for `--with-capability`) and no `CHANGELOG.md` entry — checklist is still fully unchecked.

omnibenchmark/cli/run.py
```diff
@@ -951,8 +1036,73 @@ def _generate_explicit_snakefile(
...
+            stage
+            for stage in topo_sorted_stages
+            if topo_index.get(stage.id, len(topo_index)) <= target_pos
+        ]
+        logger.info(
+            f"Module mode: expanding {len(stages_to_expand)} stage(s) "
+            f"(up to and including '{target_stage.id}'), "
+            "first expansion only."
+        )
+    elif until_stage is not None:
```
DanInci, line 1152: Worth checking the interaction with #289's diamond-join guard: `detect_diamond_input_joins()` (model/validation.py:166) runs unconditionally in `core/execution.py:87`, which happens in `run.py` before this pruning ever executes. 

It validates the *whole* model, not just `until_stage`'s ancestry. So a diamond join in some unrelated stage (not an ancestor of `until_stage`) will still fail validation even though the subgraph `--until` actually needs might be perfectly valid.

Also, worth adding a test scenario that plays this out, especially after the rebase.

tests/cli/test_run_unit.py
```diff
@@ -1223,3 +1226,221 @@ def test_run_benchmark_remote_storage_true_bool_appended(tmp_path):
...
+    def test_unrelated_branch_is_pruned(self):
+        # root feeds both branch_a and branch_b; --until branch_a drops branch_b
+        # even though it is declared before branch_a.
+        parents = {"branch_a": {"root"}, "branch_b": {"root"}}
+        result = _apply_until_filter(
+            self._stages("root", "branch_b", "branch_a"), "branch_a", parents
+        )
+        assert [s.id for s in result] == ["root", "branch_a"]
+
+    def test_diamond_keeps_both_paths(self):
```
DanInci, line 1413: This models `d` with two divergent direct parents (`b`, `c`) — exactly the shape `detect_diamond_input_joins` rejects at `ob validate`/`ob run` time. In a real benchmark this topology would never reach `_apply_until_filter` at all. 

Not wrong as an isolated algorithm test, but worth a comment noting it's not a reachable real-world case.

DanInci (commented):
Also, worth adding `CHANGELOG.md` entry and update the docs?

omnibenchmark/cli/run.py
btraven00: done

omnibenchmark/cli/run.py
btraven00: Good catch; to me, it still makes sense to "validate whole, then cut", esp. if the diamond-guard is a stopgap measure until gather lands (possibly in this release cycle).

still, good to have it documented. adding a test.

tests/cli/test_run_unit.py
btraven00: added comment

btraven00:
added missing CHANGELOG & docs; addressed comments in review.

DanInci (approved):

Issue #360, opened 2026-06-29
filtering of the execution graph defined in a master plan


btraven00:
As real benchmarks grow in size, it becomes cumbersome to edit manually parts of this plan, and keep subsets in sync (example, what we're calling "slicing").

Part of this can be solved by automation (see, for instance, the chatops in omni-scrna benchmark, powered by [obflow](https://github.com/btraven00/obflow). 

We reason that, at least some parts of this filtering can become natively supported by omnibenchmark.

What do we understand by filtering can be very broad, so I'm trying to do some distinctions here. 

- Filter a plan to just a set of stages, modules and parameters. This can be defined by order (i.e., first module by convention can be taken as the one providing test data), host capabilities (i.e. slice only the modules that do not need a GPU) or by properties of the modules themselves (the clearer example is data: some datasets carry annotations that make them suitable to feed a set of modules in different stages later on).

- Trim a benchmark to execute just up to a given stage, with different degrees of resolution (i.e. can be execute up to this stage, and then expand all the parameter grid for this module; or perhaps execute test data up to this stage and shallow-expand parameters for all modules in the stage).

- Allow for a repeated application of custom filters, even if the parent plan drifts. For complex chains, the user might be interested in defining a set of filters (combination of stage/module/param filters) and apply them to an evolving plan (obeditor does something like this currently).

btraven00:
closing since we approved #353

PR #353, opened 2026-06-15
feat: filtering design doc 

btraven00:
## Ticket

This is the first of a series of PRs that addresses partial execution and filtering of the benchmark plan.

For a general discussion, including the idea of reusable slicing #360

For conditional execution: #331

## Description

Scope is the design document 008 for module filtering and expansion

This PR is to have consensus/approval of the features that will be landed in successive PRs:

* [ ] #362  filtering by host capabilities (example: `--with capabilities gpu`)
* [ ] #361 `--until STAGE` flag
* [ ] #354 filtering by module ancestors
* [ ] #363 ob run --filter to apply reusable filters

## Checklist:
- [x] My code follows the code style of this project.
- [x] My CLI method respects the signature defined in the task.
- [x] I have documented the CLI method accordingly.
- [x] I have added tests to cover my changes.
- [x] I have added a CHANGELOG entry.

## Remarks for the reviewer:

Split from #332  - In here we land just the design doc. See the list above for related PRs.

PR #362, opened 2026-06-29
feat: capability gating

btraven00:
## Ticket

Covered/discussed in #360 

Spec in #353 

## Description

Add filtering by host capability. Provided at execution time with `--with-capability NAME`

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This document records all notable changes to `omnibenchmark`.
 This project adheres to [Semantic Versioning](https://semver.org/) and [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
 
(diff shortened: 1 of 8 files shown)


DanInci (commented):
Two things before merge:

1. **Missing CHANGELOG entry.**

2. **Prune-by-default is right, but the failure mode is silent.** Defaulting to "no capabilities present" (rather than run-and-fail) makes sense for reproducibility. But forgetting `--with-capability gpu` on a machine that has one silently drops that module. Consider a one-line warning log at start or end of the run ("N module(s) pruned by capability gate, rerun with --with-capability X") so it's not buried in per-module INFO logs.

btraven00:
rebased, added CHANGELOG and logging for pruned modules

DanInci (approved):

PR #363, opened 2026-06-30
feat(obfilter): ob run --filter applies obeditor picks to prune the DAG

btraven00:
Add --filter flag to ob run. This does runtime filtering of the plan according to a passed blob or .obfilter file (generated by obeditor).

- ob run --filter FILE|BLOB - the blob unpacks the representation of filters.
- Incomplatible with -m (module filter)
- Does drift-check comparing the summary_hash of the plan
- If we pick modules that do not exist, fails unless --allow-drift is provided.
- --dry outputs the filtered Snakefile as expected
- Tests: 23 unit + 2 e2e (prune, orphan-fail).

## Ticket

#360 for the general discussion of filtering

## Checklist:

## Remarks for the reviewer:

### DIFF
--- a/omnibenchmark/cli/run.py
+++ b/omnibenchmark/cli/run.py
@@ -138,6 +138,27 @@ def format_pydantic_errors(e: PydanticValidationError) -> str:
     default=None,
     help="Write telemetry to file instead of stdout. Allows Rich progress to remain active. Implies --telemetry.",
 )
+@click.option(
+    "--filter",
+    "filter_arg",
+    default=None,
+    metavar="FILE|BLOB",
+    help=(
+        "Run a slice of the benchmark defined by an obfilter selection: a path to a "
+        "file containing a packed blob, or an inline packed blob. Prunes the DAG to "
+        "the picked stages/modules/param-combos. Incompatible with -m/--module."
+    ),
+)
+@click.option(
+    "--allow-drift",
+    is_flag=True,
+    default=False,
+    help=(
+        "With --filter: if the benchmark has changed since the filter was made and "
+        "some selected stages/modules/parameters no longer exist, run the part that "
+        "still matches instead of erroring. Skipped selections are listed as warnings."
+    ),
+)
 @click.argument("snakemake_args", nargs=-1, type=click.UNPROCESSED)
 @click.pass_context
 def run(
@@ -154,6 +175,8 @@ def run(
     module_filter,
     telemetry,
     telemetry_output,
+    filter_arg,
+    allow_drift,
     snakemake_args,
 ):
     """Run a benchmark.
@@ -198,6 +221,27 @@ def run(
             f"-m {module_filter}: single execution path, first param expansion only."
         )
 
+    if module_filter and filter_arg:
+        log_error_and_quit(
+            logger,
+            "--filter and -m/--module cannot be combined: both define the execution "
+            "selection (-m is dev mode, --filter is a curated slice).",
+        )
+        return
+
+    filter_blob = None
+    if filter_arg:
+        from omnibenchmark.filter import unpack_blob, FilterError
+
+        packed = filter_arg
+        if Path(filter_arg).exists():
+            packed = Path(filter_arg).read_text().strip()
+        try:
+            filter_blob = unpack_blob(packed)
+        except FilterError as e:
+            log_error_and_quit(logger, f"--filter: {e}")
+            return
+
     _run_benchmark(
         benchmark_path=benchmark,
         cores=cores,
@@ -211,6 +255,8 @@ def run(
         module_filter=module_filter,
         telemetry=telemetry,
         telemetry_output=telemetry_output,
+        filter_blob=filter_blob,
+        allow_drift=allow_drift,
         snakemake_args=list(snakemake_args),
     )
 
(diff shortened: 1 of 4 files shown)


DanInci (commented):
**Stale base.** Needs a rebase.

**Docs/process.** No `howto.md` section, no `CHANGELOG.md` entry, checklist unchecked. Same as the sibling PRs.

**Integration risk.** Third open PR (#361, #362, #363) editing the same module-selection block in `_generate_explicit_snakefile`. These will conflict and need one reconciled precedence order for `-m`, `--until`, `--with-capability`, `--filter`.

omnibenchmark/cli/run.py
```diff
@@ -192,6 +215,27 @@ def run(
...
+            "selection (-m is dev mode, --filter is a curated slice).",
+        )
+        return
+
+    filter_blob = None
+    if filter_arg:
+        from omnibenchmark.filter import unpack_blob, FilterError
+
+        packed = filter_arg
+        if Path(filter_arg).exists():
```
DanInci, line 237: Verified crash: for a large enough inline blob, `Path(filter_arg).exists()` raises `OSError: File name too long` instead of returning `False`.

omnibenchmark/cli/run.py
```diff
@@ -1190,6 +1275,10 @@ def resolve_module_task(cache_key, module, software_env_id):
                 modules_to_expand = [m for m in stage.modules if m.id == module_filter]
             else:
                 modules_to_expand = stage.modules[:1]
+        elif picks is not None:
```
DanInci, line 1250: Unlike `--until` (#361), which drops metric collectors referencing pruned stages (`_filter_collectors_by_stages`), `--filter` has no equivalent. `resolve_metric_collectors` gathers all outputs from resolved_nodes of referenced stages unconditionally, so a fully-pruned stage leaves a dependent collector with an empty input set or raises. Likely needs the same drop-with-warning treatment `--until` has.

omnibenchmark/filter.py
```diff
@@ -0,0 +1,156 @@
...
+    orphans = []
+    stages_by_id = {s.id: s for s in model.stages}
+    for stage_id, mods in picks.items():
+        stage = stages_by_id.get(stage_id)
+        if stage is None:
+            orphans.append(f"stage '{stage_id}' (not in benchmark)")
+            continue
+        modules_by_id = {m.id: m for m in stage.modules}
+        for key, spec in mods.items():
+            if key == WILDCARD:
```
DanInci, line 140: This skips drift-checking for any wildcard entry, including `"*": ["some_hash"]`, which is a valid pick per `validate_picks`. If that hash matches nothing, `filter_params` returns `[]` for every module and the stage silently produces zero nodes with no orphan warning. Narrow case, but silent.

omnibenchmark/filter.py
```diff
@@ -0,0 +1,156 @@
...
+    packed = base64.urlsafe_b64encode(gzip.compress(raw, compresslevel=9, mtime=0))
+    return packed.decode("ascii").rstrip("=")
+
+
+def unpack_blob(packed):
+    """Inverse of pack_blob. Returns the blob dict; raises FilterError on bad input/version."""
+    packed = packed.strip()
+    pad = "=" * (-len(packed) % 4)
+    try:
+        raw = gzip.decompress(base64.urlsafe_b64decode(packed + pad))
```
DanInci, line 47: Decompression-bomb risk: no size limit before `json.loads` runs. Since this blob is meant to travel between people (`obeditor`-generated, pasted or shared as a file), that's a real trust boundary. 

Suggest bounding decompression itself (e.g. `zlib.decompressobj()` in a loop with a `max_length` cap, bailing with `FilterError` past a few MB) rather than calling `gzip.decompress()` unbounded.

DanInci (commented):
One more finding:

omnibenchmark/cli/run.py
btraven00: Fixed in 53eac1e: an `OSError` from the path probe now falls through to treating the argument as an inline blob. Covered by `test_run_filter_accepts_long_inline_blob`.

omnibenchmark/cli/run.py
btraven00: The collector drop already applied here: collectors are filtered on stages that actually produced nodes, which includes stages emptied by `--filter`. 4ad1ca9 makes the message name `--filter` as the cause and logs it at info level, since the slice was requested.

omnibenchmark/filter.py
btraven00: Fixed in 00b6d87: a `"*": [hash, ...]` pick is now drift-checked, and a hash no module in the stage expands to is reported as an orphan (so it fails without `--allow-drift`). Tests: `test_wildcard_hash_orphans_when_no_module_expands_it`, `test_wildcard_hash_satisfied_by_any_module_in_stage`.

omnibenchmark/filter.py
btraven00: Fixed in f8b765f: decompression goes through `zlib.decompressobj` with a `max_length` cap (`MAX_BLOB_BYTES`, 4 MB) and raises `FilterError` past it. Covered by `test_unpack_rejects_decompression_bomb`.

btraven00:
Rebased onto main (0d73ba2), which already has #360 and #361. Since main moved stage expansion into `core/_expand.py`, the `--filter` hooks now live in `expand_scatter_stage` and `expand_gather_stage`.

58ddd69 adds the `howto.md` section and the CHANGELOG entry, and documents how the flags combine: `--filter` cannot be combined with `-m/--module`, and it bypasses the `--with-capability` gate.

PR #368, opened 2026-07-09
feat: reject excludes that leave a module with no valid lineage

DanInci:
## Description

Adds a static validation that catches a class of `exclude` misconfiguration which currently passes `ob validate plan` but fails at run time with the opaque `Could not resolve input <id>`.

**The problem.** `exclude` makes two module ids mutually incompatible within a lineage. When a stage's excludes — or those of the stages producing its inputs — eliminate *every* root ("initial"-stage) lineage that could supply its inputs, that module expands to **zero nodes**. Its outputs are then never produced, and every downstream consumer fails to resolve them. Stage-level topology validation can't see this because it doesn't simulate `exclude`-based node pruning.

Concrete example (the shape that surfaced this — see `benchmark_exclude_zero_nodes.yaml`): `DATA` offers two collections, and `FEAT`/`PCA` are pinned to *different* ones, while `PCA` depends on `FEAT`:

| Stage/module | `exclude` | ⇒ runs only on | needs |
|---|---|---|---|
| `FEAT/fe` | `dB` | `dA` | `raw` |
| `PCA/pc` | `dA` | `dB` | `feat.out` (from `FEAT`) |

`PCA` can only run on `dB`, but `feat.out` is only produced on `dA` → `PCA` produces no nodes → consumers hit `Could not resolve input`.

**The fix.** `BenchmarkValidator.detect_unsatisfiable_excludes()` propagates, per module in topological order, the set of root modules from which a valid `exclude`-compatible lineage exists that satisfies all its inputs. A module with an empty set can never run and is reported. The set is an over-approximation (it doesn't track pairwise incompatibilities between two *different* producers on one lineage), so an empty set is a **sound** signal — no false positives. It runs in `validate_execution_context`, so `ob validate plan`/`ob run` enforce it while model loading for inspection/rendering (mermaid/dot, `ob describe`) stays permissive.

Only **primary** causes are reported (downstream cascades suppressed): on the plan that triggered this, the output drops from 16 errors to the 2 real ones, e.g.

> Module 'PCA/pc' can never run: its `exclude` rules (with those of its input-producing modules) leave no valid upstream lineage, so it expands to zero nodes and downstream consumers fail with 'Could not resolve input'. Input 'feat.out' (produced by stage 'FEAT') is only available on lineages excluded here. Reconcile the `exclude` lists on this module and the stages producing its inputs. See omnibenchmark/omnibenchmark#289.

## Checklist:
- [x] My CLI method respects the signature defined in the task. _(n/a — no new CLI command; new validation behaviour on `ob validate plan`)_
- [x] I have documented the CLI method accordingly. _(n/a — surfaced via the validation error message)_

## Remarks for the reviewer:

- `benchmark_exclude_zero_nodes.yaml` is a minimal, abstract repro. Tests: `tests/cli/test_validate.py::TestValidatePlanCLI::test_validate_plan_rejects_exclude_zero_nodes` (CLI) and `tests/model/test_exclude_reachability.py` (unit — flagged case **and** a benign `exclude` case that must pass, guarding against false positives; `Benchmark_001`'s existing `exclude`s also still validate clean).
- Soundness note: the reachable-root set over-approximates, so the check only ever *misses* some zero-node cases (false negatives) — it never rejects a valid plan.
- No new run-time behaviour: purely additive validation, gated to the execution-context path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/) and [Conventi
 - fix: properly render edges among stages based on I/O
 - fix: expand stages in topological order so a stage declared before an upstream-only producer resolves instead of failing with an opaque `Could not resolve input <id>` at run time (#289)
 - feat: `ob validate plan` rejects "diamond" input joins — a stage collecting inputs from two divergent branches (neither upstream of the other) — with an actionable error that names both branch stages, instead of surfacing only as a runtime warning (#289)
+- feat: `ob validate plan` rejects modules whose `exclude` rules leave no valid upstream lineage — they would expand to zero nodes and make downstream consumers fail with `Could not resolve input` at run time; the error names the module that can never run and the input it can no longer receive (#289)
 - fix(pkg): give all authors emails so PyPI lists them correctly instead of partitioning name-only entries into the Author field
 
 ## [0.5.3](https://github.com/omnibenchmark/omnibenchmark/releases/tag/v0.5.3) (Jun 15th 2026)
--- a/omnibenchmark/model/benchmark.py
+++ b/omnibenchmark/model/benchmark.py
@@ -1315,6 +1315,8 @@ class Benchmark(DescribableEntity, BenchmarkValidator):
 
         errors.extend(self.detect_diamond_input_joins())
 
+        errors.extend(self.detect_unsatisfiable_excludes())
+
         # Validate software environment paths (if benchmark_dir provided)
         if self.software_backend != SoftwareBackendEnum.host:
             for env in self.software_environments:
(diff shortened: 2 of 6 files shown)


btraven00:
tbd: rebase and check failing tests

btraven00:
noting here that this is a stop gap for more expressive filtering by lineage https://github.com/omnibenchmark/omnibenchmark/pull/354 

it's probably good to leave excludes as a quick syntax for simple needs and perhaps indicate in the documentation that provides/requires is preferred for more complex topologies.

Also scheduling for a patch release 0.5.4

btraven00 (approved):
🐋

btraven00:
During the omni-scrna hackathon, this was identified as a clear blocker for the existing topologies. 

Rebased, squashed and force-pushed.

PR #370, opened 2026-07-15
fix: replace O(2^depth) ancestor recursion in Snakefile generation

btraven00:
Node expansion built each node's input map by collecting its lineage ancestors with get_ancestor_nodes(), which found ancestors by scanning all resolved_nodes for id-prefix matches and then RE-RECURSED on every match with no memoization. For a lineage of depth k that is T(k)=Σ T(i) = O(2^k), each level also doing an O(N) scan -- so generation hung outright on deep benchmarks (30-module scRNA plan never finished; the faulthandler stack sat in get_ancestor_nodes at run.py:1343).

The bug has existed since explicit Snakefile generation landed, but only bites when lineages actually get deep. Topological stage expansion (#289, 8f21934) is what surfaced it: expanding stages in dependency order lets every producer resolve before its consumers, so the 18-stage plans now chain into deep linear lineages instead of stalling out at shallow depth in declaration order. 2^depth then became fatal.

Node ids are built as `{parent.id}-...`, so the prefix-ancestor set is exactly the parent_id chain -- a linked list. Walk it directly (O(depth), no scan), the same walk _lineage_module_ids already uses. Shallowest-first so the deepest ancestor still wins on any output-id collision.

Two adjacent O(N) scans in the same loop fixed while here:
- _select_input_nodes: next(n for n in resolved_nodes ...) -> nodes_by_id O(1) lookup (2.6s of a 30-module run).
- get_output_ids_for_node: per-call stage scan + pydantic list.index() -> precomputed stage_id->output_ids map (~180k __eq__ calls removed).

30-module scRNA plan: hang -> 6.7s, identical 26197 rules.

### DIFF
--- a/tests/cli/test_run_unit.py
+++ b/tests/cli/test_run_unit.py
@@ -16,10 +16,13 @@ from omnibenchmark.cli.run import (
     _run_snakemake,
     _substitute_params_in_path,
     _build_template_context,
-    _select_input_nodes,
-    _satisfies_requires,
-    _lineage_module_ids,
     run,
+)
+from omnibenchmark.core._lineage import (
+    select_input_nodes,
+    satisfies_requires,
+    lineage_module_ids,
+    lineage_ancestors,
 )
 from omnibenchmark.core.prefetch import populate_git_cache
 from omnibenchmark.model import SoftwareBackendEnum
@@ -299,7 +302,7 @@ class TestBuildTemplateContext:
 
 
 # ---------------------------------------------------------------------------
-# _satisfies_requires
+# satisfies_requires
 # ---------------------------------------------------------------------------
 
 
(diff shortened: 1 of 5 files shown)


markrobinsonuzh:
This has been running well for me.

markrobinsonuzh (approved):

tests/cli/test_run_unit.py
```diff
@@ -426,14 +428,61 @@ def test_dashed_module_ids_resolve_via_parent_links(self):
...
+        by_id = _by_id(download, prep, method)
+        assert lineage_ancestors(method, by_id) == [download, prep]
+
+    def test_only_own_chain_not_sibling_branch(self):
+        download = _make_lineage_node("adamson", node_id="d")
+        sibling = _make_lineage_node("norman", node_id="n")  # different root
+        child = _make_lineage_node("prep", parent_id="d", node_id="d-p")
+        assert lineage_ancestors(child, _by_id(download, sibling, child)) == [download]
+
+    def test_deep_chain_is_linear_not_exponential(self):
```
DanInci, line 463: `pytest-timeout` is already a dev dep, so `@pytest.mark.timeout(5)` here would turn the hang into a clean failure instead of a wedged CI job.

DanInci (commented):
Nice catch on the root cause. Verified the prefix/parent-chain equivalence holds (node ids are built as `{parent.id}-...` at run.py:1306, parent_id set at run.py:1419).

DanInci (approved):

tests/cli/test_run_unit.py
btraven00: added, thanks!

PR #376, opened 2026-08-14
feat: fan-in with joins and gather (#289)

### LINKED ISSUE #289: support for multi-stage outputs

btraven00:
Makes fan-in expressible. A benchmark can now consume outputs from several
upstream nodes at once, whether they are alternatives, branches to join, or a
set to collect.

Closes #289.

## What changes

Three ways a stage can reference outputs produced by more than one upstream
stage. Each was broken differently before:

| producers declare | consumer declares | before | after |
|---|---|---|---|
| `method_a` → `outputs: [{id: clustering}]`<br>`method_b` → `outputs: [{id: clustering}]` | `inputs: [clustering]` | **accepted and silently wrong** — one consumer node; `method_b` ran and nothing consumed its output | one consumer node **per producer** |
| `method_a` → `outputs: [{id: a.out}]`<br>`method_b` → `outputs: [{id: b.out}]`<br>on divergent branches | `inputs: [a.out, b.out]` | **rejected at load**: *"…on divergent branches… not supported yet"* | one node with **two parents** (a join) |
| any stages declaring `clustering` | `gather: [{from: clustering}]` | **not expressible** | one node **per group**, members passed as a list |

`inputs:` syntax is unchanged in the first two rows. What decides the outcome is
whether the producers declare the **same** output id: same id means the files
are interchangeable, so the consumer runs once per producer; different ids mean
the consumer needs both at once, so they join into one node.

The first row is a resolver fix and is **not api-gated** — with a single
producer per id the plan is byte-identical, so the only benchmarks whose plans
change are the ones that were already losing work. The join and `gather` are
gated on **api 0.7.0**.

## Gather

```yaml
- id: metrics
  gather:
    - from: clustering            # every producer of this id
      group_by: data              # one node per ancestor `data` module
  outputs:
    - id: metrics.summary
      path: "{data}_summary.tsv"
```

`group_by` is optional — omitted, every producer collects into one node, which
is what a `metric_collector` is. The cut chain roots at the stage id
(`metrics/<group>/<module>/<params>/`), so there is no path field to declare.
Gather nodes are ordinary nodes: downstream stages chain off them normally.

## Other behaviour changes

- **`provides` labels are owned by one stage.** A second stage declaring the
  same label is now a parse-time error. This is what makes a join's label union
  well-defined, and it removes an undesigned silent overwrite on plain chains.
- **A join inherits labels from every branch**, so `requires:` downstream of a
  join can name a label from any of them. `exclude` already worked this way; the
  two agree now.
- **Stage-level topology drew one edge per output id** regardless of how many
  stages produced it, so a diagram showed a consumer depending on one of its
  producers. Fixed in `get_stages_by_output`, which mermaid, dot and obeditor
  all reach through `stage_adjacency`.

## Provenance

A fan-in node's path no longer *is* its lineage — a gather's records only the
group key, a join's only its deepest input. Two additions cover that: the
directory segment carries a digest of the parent set (without it two joins
sharing a branch collide, and Snakemake raises `AmbiguousRuleException`), and
every fan-in node writes a `lineage.json` sidecar naming each contributing node.
The sidecar is written before the module runs, so a gather module can read it to
label its output.

## Also in here

- The planner moved out of `cli/run.py` into `core/_expand.py`,
  `core/_prune.py` and `core/_lineage.py`. `run.py` drops 2195 → 1313 lines and
  keeps the CLI command, the Snakemake subprocess and the resolve/report loop.
  No behaviour change.
- Design docs updated to match: 004 v6 (the YAML surface), 007 v2, 010 v3.

## Testing

Full suite green. New e2e fixture runs scatter → gather → scatter for real and
checks the aggregate against what the upstream modules computed, so collecting
the wrong members fails, not just the wrong number. Unit coverage for
membership, grouping, bundle selection and topology edges — all now marked
`short`, since most of the gather tests were unmarked and `tox -m short` was
skipping them.


omnibenchmark/core/_lineage.py
```diff
@@ -62,53 +158,286 @@ def select_input_nodes(
...
+            for input_id in declared_input_ids
+            if not any(p.id in anchor_ids for p in producers.get(input_id, []))
+        ]
+        if not missing:
+            bundles.append((anchor,))  # linear / covered — fast path
+            continue
+        # Join: for each uncovered input, take lineage-consistent partners.
+        partner_lists = []
+        for input_id in missing:
+            partners = [
```
DanInci, line 338: **1. Join partners bypass the shadowing filter** (confirmed, blocker)

`select_input_nodes` drops shadowed producers via `_maximal_producer_stages`. Partner selection here goes straight to `producers[input_id]`, filtered only by lineage consistency. Two producers of the same id on one chain are consistent with the anchor and with each other, so `product(*partner_lists)` emits a bundle per producer.

Repro on this branch. Branch B declares `x.out` at `b1` and again at its descendant `b2`; anchor `a` is deeper in topo order and declares `a.out`; consumer declares `inputs: [a.out, x.out]`:

```
select_input_bundles(...) ->
  ['a', 'b1']
  ['a', 'b2']
```

Per 010 3.1 `b1` is shadowed by `b2`, so only `['a','b2']` should exist. The extra node runs the module against the upstream file.

Suggested fix: route `producers[input_id]` through `_maximal_producer_stages` before building `partner_lists`.

omnibenchmark/backend/snakemake.py
```diff
@@ -367,6 +449,9 @@ def _write_gather_shell(self, f: TextIO, node: ResolvedNode):
             "OUTPUT_DIR=$(cd {params.output_dir} && pwd)",
         ]
 
+        # A gather cuts the chain, so its members exist nowhere in the path.
+        lines += self._lineage_sidecar_lines(node)
```
DanInci, line 501: **2. Gather file order diverges from `lineage.json` past 10 members** (confirmed)

The `sorted()` predates this PR, but gather is what makes it reachable at scale, so flagging it here. `_write_gather_shell` at ~L442 does `for key in sorted(node.inputs.keys())`, lexicographic over `input_0 .. input_N`.

`benchmark_pca_cubes.yaml` gathers 18 members per SENS node, so the Snakefile emits:

```
for _f in {input.input_0} {input.input_1} {input.input_10} {input.input_11} ... {input.input_15} {input.input_2} ...
```

while the `lineage.json` written two lines above, from `gathered_from`, is in plan order. 010 3.5 promises plan order and 3.7 pitches the sidecar as how a report module labels its N inputs, so a module pairing the i-th path with the i-th member mislabels everything past the 10th. Still deterministic, so no rerun churn.

Suggested fix: sort numerically.

omnibenchmark/model/benchmark.py
```diff
@@ -658,6 +696,44 @@ class Stage(DescribableEntity):
...
+    prefix: Optional[str] = Field(
+        None,
+        description=(
+            "Filesystem prefix for a gather stage's cut chain (design 010 "
+            "§3.3). Required when `gather` is set; ignored otherwise."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def validate_gather(self) -> "Stage":
```
DanInci, line 717: **3. A gather stage is classified as an initial stage** (confirmed)

`is_initial` (~L1412) is `stage.inputs is None or len(stage.inputs) == 0`, and a gather replaces `inputs:` with `gather:`. `get_stage_implicit_inputs` was taught about gather in this PR; `is_initial` wasn't, so every gather stage now reads as a root.

The pre-existing initial-stage `requires` check then fires on gather modules. Adding `requires: {kind: a}` to the gather module in `tests/e2e/configs/13_gather.yaml`:

```
Module 'MC' in initial stage 'metrics' declares `requires`, but an initial stage
has no upstream lineage to match against. Remove the `requires` gate or move the
module to a later stage.
```

Rejecting is right, since `expand_gather_stage` never evaluates `module.requires`. But both halves of the message are wrong for a gather: it has upstream lineage, and moving to a later stage doesn't help. A gather-specific message pointing at 010 3.3 would be accurate.

Same classification also gives a gather `input_dirname = out_dir` in `core/_graph.py:25`, and `get_dataset_ids` (`_graph.py:332`) breaks on the first initial stage, so a gather declared before the data stage would be reported as the dataset stage. `SENS` and `REPORT` in `benchmark_pca_cubes.yaml` both report `is_initial=True`.

omnibenchmark/core/_lineage.py
```diff
@@ -62,53 +158,286 @@ def select_input_nodes(
...
-        node = nodes_by_id.get(node.parent_id) if node.parent_id else None
+    lineage = {input_node.module_id}
+    for ancestor in iter_ancestors(input_node, nodes_by_id, through_gather=False):
+        lineage.add(ancestor.module_id)
     return lineage
 
 
-def lineage_ancestors(input_node, nodes_by_id) -> list:
-    """Strict ancestors of ``input_node``, shallowest first.
+def ancestor_module_at_stage(member_id, group_stage, nodes_by_id):
```
DanInci, line 217: **4. Gather group key ignores `Module.provides`** (plausible, narrow)

`ancestor_module_at_stage` returns the raw `module_id`, while `resolve_label_value` (L366) gives pre-gather nodes `module_provides[label]` when bound. Same label name, two values across the cut:

- data stage `provides: [data]`, module `D1` with `provides: {data: custom}`
- upstream nodes carry `data=custom`
- the gather with `group_by: data` binds `data=D1`
- a downstream `requires: {data: custom}` matches before the gather, prunes silently after

Probably deferred to Phase 2. If so, 010 3.3 should say the group value is always the ancestor module id, never the bound label value. Right now 3.2 and 3.3 read as if they agree.

omnibenchmark/backend/snakemake.py
```diff
@@ -68,9 +84,11 @@ def __init__(
         self.benchmark_version = benchmark_version
         self.benchmark_author = benchmark_author
         self.api_version = api_version
+        self._nodes_by_id: dict = {}
```
DanInci: **5. `lineage.json` degrades silently when `_nodes_by_id` is unpopulated** (plausible, low)

`_nodes_by_id` is only assigned in `generate_snakefile`. Any caller reaching `_write_node_rule` another way gets `_member_record` falling back to a bare `{"node_id": ...}`: no stage, module, commit, params or dir, and no diagnostic. That is the provenance the sidecar exists to preserve.

`DebugSnakemakeGenerator` overrides the shell writers and could plausibly be driven per-node. Passing the node map into the writer, or asserting it's populated before emitting a fan-in rule, would make it loud.

DanInci (commented):
## Review: fan-in with joins and gather (#289)

Read the diff, ran the planner over `benchmark_pca_cubes.yaml` with stubbed module resolution, and generated its Snakefile. Unit tests pass.

Plan is correct on that example: 5 DATA, 10 PREP, 80 PCA, 5 grouped SENS (18 members each), 1 global REPORT, no DAG errors. `stage_adjacency` picks up `PREP -> SENS ['truth_tsv']`, i.e. a gather dependency never declared under `inputs:`.

Ten inline comments, five on source and five on `docs/design/010-gather.md`. Finding 1 is the only one I'd block on.

docs/design/010-gather.md
```diff
@@ -0,0 +1,470 @@
...
+<entrypoint> --output_dir … --name summarizer \
+    --clustering /abs/p1.tsv /abs/p2.tsv /abs/p3.tsv \
+    [param flags]
+```
+
+Output handling is unchanged from any other node and follows whatever the
+declared api version specifies — the api ≥ 0.5 contract today, design 009's
+named outputs when it lands. There is no gather-specific rule.
+
+Member order is plan order: topological by stage, then declaration order within
```
DanInci: **§3.5 specifies ordering where it means determinism, and the code follows neither half**

> Member order is plan order: topological by stage, then declaration order within a stage, then parameter expansion. Arbitrary but deterministic [...] Modules must not read meaning from position; identity comes from the paths.

These two sentences pull against each other. If position carries no meaning, specifying the order serves no purpose. The reason actually given is byte-identical Snakefiles, which is a stability requirement, not an ordering one.

The implementation then satisfies the requirement that matters (deterministic) and violates the one that got written down. `_write_gather_shell` sorts the enumerated keys lexicographically, so past 10 members the CLI order is `input_0, input_1, input_10, input_11, ... input_2` while `lineage.json` stays in plan order. See my inline comment on `backend/snakemake.py`.

Suggest rewording to "deterministic and stable across runs; position carries no meaning". That is honest, implementable, and would have made the sort a non-issue instead of a spec violation.

docs/design/010-gather.md
```diff
@@ -0,0 +1,470 @@
...
+`group_by` is optional. Omitted, every producer lands in one node and the
+output path carries no group segment — that is the global form, and it is what
+a `metric_collector` is (§3.6).
+
+Grouping merges parameter expansions: members descending from `data-D1.p1` and
+`data-D1.p2` land in the same group, because the key is the ancestor **module**
+id (`D1`), not its node id. Per-parameter grouping needs a label-based
+`group_by` (§6, Phase 2).
+
+**Deferred**: `where` (member filtering), label-value `group_by`, and plain
```
DanInci, line 213: **§3.2: multiple gather entries with partially populated groups are undefined**

`gather: [{from: A, group_by: D}, {from: B, group_by: D}]` is legal (one axis, so `validate_gather` allows it) and `benchmark_pca_cubes.yaml` uses exactly that shape. The spec covers a group that is empty, but not a group populated for one entry and empty for another.

Built the case, two gather entries with one branch excluded for one dataset:

```
metrics-MC-D1.default: members=2 flags=['methods.other', 'methods.result']
metrics-MC-D2.default: members=1 flags=['methods.result']
```

The D2 node is created with `--methods.other` simply absent. No warning. The module fails at runtime with an argparse error, a long way from the cause.

§2 lists "No silent absence: every dropped member and empty group is logged" as a goal, but this is neither a dropped member nor an empty group, so no rule reaches it. §3.2 needs a third case: is a partial group an error, a warning, or an empty flag? I'd argue error, since a module declaring two gather inputs cannot meaningfully run with one.

docs/design/010-gather.md
```diff
@@ -0,0 +1,470 @@
...
+holds the member node ids in the resolved plan; `lineage.json`, written beside
+the outputs, holds the same record on disk (§5.2). Full ancestry of any
+post-gather node is recovered by walking `parent_id` as usual and, at a gather,
+fanning out through the members.
+
+**Identity and paths.** A gather node's id is composed from the group key
+(`metrics-summarizer-D1.default`) rather than a parent chain, and `parent_id` is
+`None`. Downstream ids prefix-compose off it as usual, so chains resume
+immediately: the plan alternates scatter segments and gather cut points, never a
+general DAG. Outputs land under `<prefix>/<group>/<stage>/<module>/<param>/…`,
```
DanInci: **§3.3 introduces `prefix` with no shape rules**

`prefix` is made mandatory for a gather and is the root of its output tree, but nothing says what a valid one is. `validate_gather` only checks presence:

```
prefix '../../../etc' ACCEPTED -> ['../../../etc']
```

A design adding a user-controlled filesystem field should answer three things: is it relative to the output dir, is traversal allowed, and may two gather stages share one? Right now all three are answered by accident rather than by rule.

(The third is benign today, since `<prefix>/<group>/<stage>/...` disambiguates on stage id, but that is a property of the current layout, not a stated guarantee.)

docs/design/010-gather.md
```diff
@@ -0,0 +1,470 @@
...
+the existing nested builder resuming below the prefix.
+
+### 3.4 Membership and pruning order
+
+Membership is computed **after** all pruning:
+
+```
+excludes → requires → capability gates → gather membership → where → group_by partition
```
DanInci: **§3.4 states a normative order containing a deferred step**

```
excludes → requires → capability gates → gather membership → where → group_by partition
```

`where` is deferred per §3.2, so this pipeline documents a stage that does not exist. Harmless as a sketch, but it sits under "Membership is computed **after** all pruning", which reads as a contract. Marking the step `(deferred)` inline would keep it forward-looking without implying it runs.

DanInci (commented):
Continuation of the review above: comments on `docs/design/010-gather.md`.

docs/design/010-gather.md
DanInci: Also, what are the limitations of this? Surely one cannot escape the current root folder of the output?

DanInci: Answering my own question: no, nothing stops it. `prefix: ../../escaped` goes straight into the Snakefile, and propagates to stages chaining off the gather:

```
"../../escaped/D1/metrics/MC/.default/metrics.json"
"../../escaped/D1/metrics/MC/.default/report/R1/.13058ed0/R1_r.json"
```

Snakemake runs with the output dir as cwd, so that writes above the run root. `validate_gather` only checks presence.

Rather than validate it, I'd drop the field and root the gather at its stage id. 007 line 69 already says one top-level dir per stage, and stage ids are unique, so `<stage>/<group>/<module>/<param>/` needs no prefix. The example shows the redundancy: `prefix: report` on stage REPORT gives `report/REPORT/quarto/...`.

btraven00: agree, good simplification, let's use the stage

btraven00: done, pushed on commit below

docs/design/010-gather.md
btraven00: fixed, turned into an error

docs/design/010-gather.md
btraven00: agree, addressed

omnibenchmark/backend/snakemake.py
btraven00: fixed the numeric sort. Left a TODO to simplify this after we land #329. Also, a reminder that we'll make our life easier if we forbid dots and other chars in the IDs. This will also unlock moving away from the 3-level depth nesting. Perhaps we can sneak a deprecation warning in `0.7`

omnibenchmark/core/_lineage.py
btraven00: addressed at 6f271f5

omnibenchmark/model/benchmark.py
btraven00: fixed at d04c5ff

omnibenchmark/core/_lineage.py
btraven00: fixed at 93dc9ec - I found a simpler solution for now is not to allow label collisions with stage names, keeping disjoint namespaces. we can refine later on.

omnibenchmark/core/_expand.py
```diff
@@ -0,0 +1,551 @@
...
+            for output_spec in stage.outputs:
+                # Only the group label is bound: a template referencing any
+                # other lineage label raises in substitute() — the plan-time
+                # error design 010 §3.3 wants, never a silent empty sub.
+                tmpl = ctx.substitute(output_spec.path, params=params)
+                group_dir = f"{gval}/" if gval is not None else ""
+                # The cut chain roots at the stage id: unique by construction,
+                # so two gathers cannot share a tree.
+                output_path = truncate_path_filename(
+                    f"{stage.id}/{group_dir}{module_id}/{param_id}/{tmpl}"
```
DanInci, line 205: **`ob collect` mis-parses every path below a grouped gather** (confirmed)

The group segment makes this the first 4-segment node in the tree (`<stage>/<group>/<module>/<param>`). Every other node is exactly 3, nested or flat. `cli/collect.py:81` is `zip(*(iter(rel_parts),) * 3)`, so everything below the cut shifts by one.

`ob collect performance` on the tree the `13_gather.yaml` e2e leaves behind:

```
stage       module    dataset   param_hash  lineage
data        D1        D1        94686c86    data/D1
method_a    MA        D1        85ff2664    data/D1/method_a/MA
.default    report    D1        R1          metrics/D1/.default/report
.default    report    D2        R1          metrics/D2/.default/report
```

Truth for the last two is `stage=report, module=R1, param_hash=13058ed0` — `stage` is a param dir, `module` is a stage id, `param_hash` is a module id. `_collect_params` then walks `metrics/D1/MC/parameters.json` and `.../report/R1/parameters.json`, neither of which exists, so `params` comes back empty as well. `_build_record` warns only when `triples` is empty, and 7 parts is not, so none of this is reported. `performances.tsv` is what `ob dashboard` consumes.

Scope, checked against the planner rather than assumed:

- **Grouped gather only.** A global gather stays at 3 segments and parses correctly (its `dataset` column reads `MC`, the gather module, which is arguably the honest answer — the cut has no single dataset).
- **Joins are unaffected**: the parent-set digest stays inside the existing param segment (`.default-6103e640`), so the chain stays a multiple of 3. Worth keeping that property.
- The gather node itself emits no row, since `_write_node_rule` excludes gather from the `benchmark:` directive — so the corruption is purely downstream.

`cli/collect.py` is untouched by this PR and `main` has no gather, so this is new here. Silent wrong rows rather than a skip or an error is the thing 008 §2 argues against.

Cheapest fix is probably to stop deriving structure from segment arity — the plan already knows each node's stage/module/param, so `collect` could read them rather than re-parse the path. Failing that, `_triples` needs to know a gather contributes four.

omnibenchmark/backend/snakemake.py
```diff
@@ -246,6 +309,64 @@ def _sanitize_rule_name(self, node_id: str) -> str:
...
+        block before bash sees it, and single quotes are escaped for the
+        surrounding shell quoting.
+        """
+        record = self._lineage_record(node)
+        if record is None:
+            return []
+        body = json.dumps(record)
+        body = body.replace("{", "{{").replace("}", "}}")
+        body = body.replace("'", "'\\''")
+        return [f"echo '{body}' > $OUTPUT_DIR/lineage.json"]
```
DanInci, line 368: **`lineage.json` is never archived** (confirmed)

It is not a declared Snakemake `output:` — `_write_node_rule` emits only `node.outputs` — and `storage/base.py:32` lists just `parameters.json` and `parameters_dict.tsv` in `extra_files_to_version_not_in_benchmark_yaml`. `get_expected_benchmark_output_files` is exactly those two sources unioned, so `ob archive` drops the sidecar.

So the one record that makes a fan-in node's provenance recoverable is the one file that does not survive archiving — and it is unrecoverable after the fact, since the whole reason it exists is that the path stopped carrying the lineage. An archived gather result would have no way to say what it was computed from.

Adding `f"{out_dir}/**/lineage.json"` to that list looks sufficient.

DanInci (commented):
Continuing the review. Findings 1-7 all check out: the shadowing filter now mirrors `select_input_nodes`, `_input_key_index` is only reachable from `_write_gather_shell` (gather/collector nodes use `input_N` exclusively, so the `int()` cannot throw), and the disjoint-namespace fix is not a compat break since `Stage.provides` is gated at api >= 0.7.0. Docs look right. Suite green locally (1111 passed), gather e2e passes.

Two new ones below, both confirmed by running the code. Finding 3 is only half fixed: the message is right, but `is_initial` still returns `True` for a gather and `core/_graph.py` still calls it, and that graph is built in `BenchmarkExecution.__init__` on every `ob run` — with the gather stage declared first, `get_benchmark_datasets` returns `['MC']`. No in-repo caller today, so non-blocking, but "legacy only" understates that the legacy graph is still constructed.

DanInci (changes requested):
Blocked only on **finding A** (`ob collect` mis-parses every path below a grouped gather) — silent wrong rows in `performances.tsv`, which `ob dashboard` reads.

**B** (`lineage.json` never archived) is worth fixing here but I would not hold the merge for it. Findings 1-7 verify clean; 3 is half fixed and non-blocking.

omnibenchmark/core/_expand.py
btraven00: Confirmed, fixed in cli/collect.py.

Didn't go via the plan (we only pass the out-dir).   _split_gather_group() pops the group key, _triples() aligns again, _collect_params() inserts it when walking the tree.

After 0.7 I was thinking to simplify the nesting into a flatter 1-folder structure, we can revisit paths and collection then.

omnibenchmark/backend/snakemake.py
btraven00: Confirmed, and it affected more:  `glob.glob` skips dot-prefixed names, and every node's param dir is dotted since v0.5 (.default, .<hash>) — so extra_files_to_version_not_in_benchmark_yaml has been matching nothing at all, and parameters.json isn't archived on main either. Added lineage.json to the list as you suggested, and swapped `glob.glob` for Path().glob in `storage/expected.py`, which matches hidden dirs. Regression test in tests/storage/test_expected.py.

DanInci (approved):

btraven00:
thanks @DanInci  for your review and patience! merged! :whale:

PR #379, opened 2026-08-19
backend: emit snakemake's threads: directive from resources.cores

btraven00:
The generated rules carried resources: cores=N but never a threads: directive, so threads stayed at snakemake's default of 1. Two consequences:

* Snakemake exports OMP_NUM_THREADS=<threads> into every shell job, so every module ran single-threaded regardless of the machine. A benchmark comparing BLAS implementations measured no difference at all, because both arms were pinned to one thread by the harness.
* The scheduler assumed each job cost one core, so --cores N launched N jobs that each spawned their own thread pool and oversubscribed the box.

Emitted only when a stage declares resources.cores. Defaulting it would change scheduling for every existing benchmark, so opting in stays deliberate.
.

## Checklist:
- [na] My CLI method respects the signature defined in the task.
- [na] I have documented the CLI method accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/) and [Conventi
 ## [0.7.0](UNRELEASED)
 
 - feat: capability filters by --with-capability (#360)
+- feat: emit threads per rule according if the module requests a number of cores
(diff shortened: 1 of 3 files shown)


atchox (approved):
looks good to me!

Issue #371, opened 2026-07-20
`ob describe status` is broken, at least for this omni-scrna


markrobinsonuzh:
```
mark in 🌐 omnibenchmark in split-stages-plan on  count-factorization [$?] via 📐 v4.5.3 via 🅒 omni-retreat via 🧚 
❯ ob describe status benchmark_conda.yaml 
Traceback (most recent call last):
  File "/home/mark/miniforge3/envs/omni-retreat/bin/ob", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mark/miniforge3/envs/omni-retreat/lib/python3.12/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #381, opened 2026-08-22
fix: status crashed on DAGs with several terminal stages (#371)

btraven00:
The last column of the exec-path table read a leaked loop variable, i.e. always the globally last stage. Any path terminating on another branch raised KeyError. Use the path's own terminal stage instead.

## Ticket

#371 

## Description

status reports break with current omni-scrna plan

## Checklist:
- [na] My CLI method respects the signature defined in the task.
- [na] I have documented the CLI method accordingly.


btraven00:
@markrobinsonuzh can you check the fix is valid? 

I see other problems with status, like the status command taking a few seconds, or the linked plan failing on the environment solving step, but the original exception should be gone.

markrobinsonuzh:
See this https://github.com/omnibenchmark/omnibenchmark/issues/371#issuecomment-5410051021

markrobinsonuzh (approved):

PR #383, opened 2026-08-25
fix(archive): exclude .snakemake and .cache from result archives

### LINKED ISSUE #324: archive stores more artifacts than needed
In this particular case, we probably should skip `snakemake` and `cache` if not using the `--software` flag

```
(omni-0.5.0) mark@rolandgarros:~/projects/omb/pipelines-plan$ ob archive -c -r benchmark.yaml
Created archive: scRNA_Benchmark_(pipelines)_1.1.tar.gz
[snip]
(omni-0.5.0) mark@rolandgarros:~/projects/omb/pipelines-plan$ ls -lh scRNA_Benchmark_\(pipelines\)_1.1.tar.gz
-rw-r--r-- 1 mark robinsonlab 11G Apr 29 13:20 'scRNA_Benchmark_(pipelines)_1.1.tar.gz'
(omni-0.5.0) mark@rolandgarros:~/projects/omb/pipelines-plan$
(omni-0.5.0) mark@rolandgarros:~/projects/omb/pipelines-plan$ cd out
[snip]
(omni-0.5.0) mark@rolandgarros:~/projects/omb/pipelines-plan/out$ du -h -d 1 . | sort -h
4.0K	./.envs
8.0K	./.conda
8.0K	./.config
16K	./.metadata
576K	./.logs
4.1M	./.modules
236M	./denet
277M	./datasets
2.1G	./.cache
9.4G	./.snakemake
12G	.
```

agu2347:
Fixes #324.

### Problem

`prepare_archive_results_local`'s catch-all scan of the results directory (`results_path.rglob("*")`) includes every file it finds under `results_dir`, with no exclusions. This sweeps in Snakemake's own internal execution state under `.snakemake/` (locks, logs, per-rule metadata) and tool/package caches under `.cache/`. Neither directory holds actual benchmark artifacts, but both can grow to many gigabytes during a real run.

As reported in the issue, running `ob archive -c -r benchmark.yaml` (without `-s`/`--software`) produced an 11G archive where `.snakemake` (9.4G) and `.cache` (2.1G) inside `out/` accounted for nearly all of it, versus a few hundred MB of actual `data`/`denet` results.

### Fix

Add an explicit exclusion in `prepare_archive_results_local`'s catch-all directory scan for any file living under a `.snakemake/` or `.cache/` subdirectory of `results_dir`. This only affects the catch-all "also include any additional files" pass — files explicitly expected via `get_expected_benchmark_output_files` are unaffected either way, and any real result subdirectory elsewhere in `results_dir` (even one that happens to be nested several levels deep) is untouched.

### Testing

Added `test_prepare_archive_results_local_excludes_snakemake_and_cache_dirs` to `tests/archive/test_archive_basics.py`, which creates a results dir with a real result file, a nested `.snakemake/metadata/` file, a nested `.cache/some_tool/` file, and a real result in an ordinary nested subdirectory (to confirm only `.snakemake`/`.cache` are special-cased, not nesting in general). It asserts the real files are included and no `.snakemake`/`.cache` file leaks in.

Full `tests/archive/` suite: 45 passed. Reverting just the exclusion logic reproduces the reported behavior exactly (the new test fails because the `.snakemake`/`.cache` junk files show up in the result list); reapplying restores a clean pass.


btraven00 (approved):

btraven00 (approved):

btraven00:
thanks for your contribution @agu2347 - this is now merged in main

